PR #49, opened 2025-01-02
feat: React Examples Components (DT-7024)

TheMooseman:
# What
Create react components for our examples(excluding layers).

# How
- Create components
- Ensure examples still work
- Apply learnings from implementations

# PR Checklist


examples/src/home.tsx
```diff
@@ -0,0 +1,26 @@
...
+            <ul>
+                <li>
+                    <a href="/dzi">Deep Zoom Image</a>
+                    <br />
+                </li>
+                <li>
+                    <a href="/omezarr">OMEZARR</a>
+                    <br />
+                </li>
+                {/* layers is not in the AC to be converted to a react component
```
lanesawyer, line 17: Let's at least let it render like it was, since there's a good Markdown page explaining how to use it in the docs still so we want it working.

lanesawyer (commented):
Routing works! Let's get the Layers app showing too, since that's an important one.

There is a type error that keeps Vite from building, easy fix though.

I'll poke at this again once it's fully baked!

examples/src/dzi/dzi-demo.tsx
```diff
@@ -42,40 +44,44 @@ const exampleSettings: DziRenderSettings = {
...
-                wheel={zoom}
-            />
-            <DziView
-                id="right"
-                dzi={exampleDzi}
-                svgOverlay={overlay.current}
-                camera={{ ...exampleSettings.camera, view }}
-                wheel={zoom}
-            />
+            <div style={{ display: 'flex', flexDirection: 'row' }}>
```
lanesawyer, line 72: Could you add some text to the page encouraging users to scroll to get the demos to appear? Alternatively, we figure out why it takes interaction to make these show up properly.

examples/src/dzi/dzi-viewer.tsx
```diff
@@ -63,6 +56,7 @@ export function DziView(props: Props) {
         }
         return () => {
             if (cnvs.current) {
+                // eslint-disable-next-line react-hooks/exhaustive-deps
```
lanesawyer, line 59: I think this is out of place? This isn't disabling the `useEffect` dependency lint issues. We also still need to actually integrate a linter soon 😅

examples/src/omezarr/omezarr-demo.tsx
```diff
@@ -0,0 +1,141 @@
...
+        setDragging(false);
+    };
+
+    // you could put this on the mouse wheel, but for this demo we'll have buttons
+    const handlePlaneIndex = (next: 1 | -1) => {
+        setPlaneIndex((prev) => prev + next);
+    };
+
+    return omezarr && settings ? (
+        <RenderServerProvider>
```
lanesawyer, line 97: This might be why the omezarr demo requires interaction like a scroll to show up. Always return the provider, then do the ternary inside it as children

lanesawyer (commented):
Could you remove `examples/public/layers.html`? Only `examples/layers.html` is needed.

Looking great, everything works! The demos don't render until you interact with the page, like a scroll event, and then they pop up. That would be nice to fix, or we can add a message on the screen for now.

examples/src/omezarr/omezarr-demo.tsx
TheMooseman: that fixes it, yep!

examples/src/dzi/dzi-demo.tsx
TheMooseman: In bkp we solve that problem with an "onComplete" callback with a hook, but lets wait to look into this deeper in another ticket

lanesawyer (approved):
LGTM

examples/src/layers/examples.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { AnnotationGridConfig } from '~/data-sources/annotation/annotation-grid';
+import type { ZarrSliceConfig } from '~/data-sources/ome-zarr/planar-slice';
+import type { ZarrSliceGridConfig } from '~/data-sources/ome-zarr/slice-grid';
+import type { ScatterplotGridConfig, ScatterPlotGridSlideConfig } from '~/data-sources/scatterplot/dynamic-grid';
+
+const slide32 = 'MQ1B9QBZFIPXQO6PETJ';
+const colorByGene: ColumnRequest = { name: '88', type: 'QUANTITATIVE' };
+const scottpoc = 'https://tissuecyte-ome-zarr-poc.s3.amazonaws.com/40_128_128/1145081396';
+
+export const examples: Record<string, any> = {
```
froyo-np, line 11:
```suggestion
export const examples = {
```

froyo-np: dont erase types with any...

examples/src/layers/examples.ts
```diff
@@ -0,0 +1,70 @@
...
+        type: 'ZarrSliceGridConfig',
+        gamut: {
+            R: { index: 0, gamut: { max: 20, min: 0 } },
+            G: { index: 1, gamut: { max: 20, min: 0 } },
+            B: { index: 2, gamut: { max: 20, min: 0 } },
+        },
+        plane: 'xy',
+        slices: 4,
+    } as ZarrSliceGridConfig,
+    ['structureAnnotation']: {
```
froyo-np, line 56: what happened here?
whats wrong with structureAnnotation: {...}

examples/src/dzi/use-dzi-image.ts
```diff
@@ -0,0 +1,80 @@
...
+                    return res.text();
+                })
+                .then((s) => decodeDzi(s, v))
+        )
+    );
+
+    callback(images);
+}
+
+export function useDziImages({ imageUrls, onComplete }: useDziImageProps) {
```
froyo-np, line 60: useDziImages and getImages are unused, stateful, clunky - lets remove them?

examples/src/omezarr/omezarr-demo.tsx
```diff
@@ -0,0 +1,145 @@
...
+        [omezarr, view, planeIndex]
+    );
+
+    useEffect(() => {
+        loadOmeZarr(demo_versa).then((v) => {
+            setOmezarr(v);
+            const size = sizeInUnits('xy', v.multiscales[0].axes, v.multiscales[0].datasets[0]);
+            if (size) {
+                console.log(size);
+                setView(Box2D.create([0, 0], [size[0], size[1]]));
```
froyo-np, line 46:
```suggestion
                setView(Box2D.create([0, 0], size));
```

examples/src/utils/use-why-did-you-update.ts
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * This is a useful debugging and performance hook
+ * It will log what parts of props caused the functional component to update.
+ *
+ * See: https://usehooks.com/useWhyDidYouUpdate/
+ *
+ * @param {string} name - name will show up in console log
+ * @param {Object} props - entire props object of the component being tested
+ */
+export function useWhyDidYouUpdate<P extends object>(name: string, props: P) {
```
froyo-np, line 21: remove this file?

examples/src/layers/examples.ts
```diff
@@ -0,0 +1,70 @@
+import type { ColumnRequest } from '~/common/loaders/scatterplot/scatterbrain-loader';
+import type { AnnotationGridConfig } from '~/data-sources/annotation/annotation-grid';
+import type { ZarrSliceConfig } from '~/data-sources/ome-zarr/planar-slice';
+import type { ZarrSliceGridConfig } from '~/data-sources/ome-zarr/slice-grid';
+import type { ScatterplotGridConfig, ScatterPlotGridSlideConfig } from '~/data-sources/scatterplot/dynamic-grid';
+
+const slide32 = 'MQ1B9QBZFIPXQO6PETJ';
```
froyo-np, line 7: ok this whole file is unused - it was nice that it was pulled out of the code for the layers example logic, but the types got a bit messed up. 
either fix the types and use it, or delete it and keep the actual values on line 745 of src/layers.ts

examples/package.json
```diff
@@ -61,6 +61,7 @@
         "lodash": "^4.17.21",
         "react": "^18.3.0",
         "react-dom": "^18.3.0",
+        "react-router": "^7.0.2",
```
froyo-np, line 64: are a handful of static, independant pages worth pulling in this dependency?

examples/src/dzi/dzi-viewer.tsx
```diff
@@ -91,20 +84,42 @@ export function DziView(props: Props) {
...
+                canvas.current.removeEventListener('wheel', handleWheel);
+            }
+        };
+    }, [onWheel]);
+
     return (
         <canvas
             id={id}
             ref={cnvs}
-            onWheel={wheel}
```
froyo-np, line 105: so this seemed to work before - why copy paste the long-winded version from bkp-client?

examples/src/omezarr/omezarr-demo.tsx
```diff
@@ -0,0 +1,145 @@
...
+            }
+        });
+    }, []);
+
+    const zoom = (e: WheelEvent) => {
```
froyo-np, line 51: I thought part of the improvements we were going to add was around the cameras - currently the dzi demo cant pan, only zoom. Perhaps we could consolidate their zoom and pan handlers and have both working on both?

examples/package.json
froyo-np: I suppose it does let us delete the kinda gross repetetative {demo-name.html} files....

froyo-np: lets keep it, seems fine

froyo-np (changes requested):
Cleanup please, consider camera event handing consolidation

examples/src/dzi/dzi-viewer.tsx
TheMooseman: I need the wheel event to enable zooming on mouse position, offsetX and offsetY don't exist on the react types for `React.WheelEvent` but do on `WheelEvent`

examples/src/omezarr/omezarr-demo.tsx
TheMooseman: I made simple functions for zoom and pan rather than wrap it in a hook just so it's more transparent for anyone reading the examples. I think it's easier to understand but I can move them into separate hooks if you think that's better.

froyo-np: not asking for hooks...
one of the learnings from integrating these components was that the hard-coded cameras in different spaces was confusing. My ask here is that you share the zoom & pan interactions (event handlers) between the two components, if possible.
right now, the dzi and omezarr components have different capabilities, but there is no reason for that to be the case, right?

examples/src/dzi/dzi-viewer.tsx
froyo-np: what?
dzi viewer doesnt use offsetX or offsetY in its zoom handler, so I'm not sure what you're talking about. It would be great if it did - this is adding a passive:false handler via a ref, rather than just setting the onWheel prop - that has to be done in BKP-client, but this is a demo and I dont think we need it.

examples/src/common/camera.ts
```diff
@@ -1,8 +1,44 @@
...
     readonly view: box2D; // a view in 'data space'
     readonly screen: vec2; // what that view projects to in display space, aka pixels
     readonly projection: 'webImage' | 'cartesian';
 };
+/**
+ * Zooms relative to your current mouse position
+ * @param view your current view
+ * @param screenSize the size of your canvas/screen
+ * @param zoomScale the scale you want to apply to your view
+ * @param mousePos the offsetX and offsetY of your mouse
```
froyo-np, line 14: none of these comments are helpful are they?

a parameter named screenSize is obviously the size of the screen - maybe give the units?
mousePos - is this in pixels, or the data space? 
view - what is this? obviously its the current view. this is a box, in dataspace, that gets mapped to the screen or canvas.

examples/src/common/camera.ts
```diff
@@ -1,8 +1,44 @@
...
+ */
+export function zoom(view: box2D, screenSize: vec2, zoomScale: number, mousePos: vec2) {
```
froyo-np, line 16: other than the unhelpful docs, these are excellent - thanks for unifying the camera logic!

examples/src/dzi/decode-dzi.ts
```diff
@@ -0,0 +1,81 @@
...
+                    },
+                };
+            }
+        }
+    }
+
+    return undefined;
+}
+
+/* Example XML
```
froyo-np, line 57: nice!

froyo-np (changes requested):
this is looking good, thanks for the cleanup. Since these are examples for others to follow, could you spend some time on the javadoc to explain things? "@param view the current view" is not helping anyone

examples/src/common/camera.ts
lanesawyer, line 1: This will be a good base to build on for pre-baked camera solutions as we make more comprehensive "out of the box" components!

lanesawyer (approved):
Everything still works, I think you hit all the main points Noah called out.

Even if there is something left over let's just ship it now so everything is in `main` as we keep iterating on it.

PR #50, opened 2025-01-30
fix: Removed assumption of annotation SVG's existence [DC-498]

Jarbuckle:
# fix: Removed assumption of annotation SVG's existence [[DC-498](https://alleninstitute.atlassian.net/browse/DC-498)]

## What
DZI images without associated annotation SVGs were failing to render. Since it is valid for DZI images to not have associated annotation SVGs, removed this assumption to prevent issues during rendering.

## How
  - Replaced `compose` (which was always created via `buildCompositor` and assumed an SVG's existence) with a variable `render` that falls back to `putImageData` when no SVG is found.

# PR Checklist

[DC-498]: https://alleninstitute.atlassian.net/browse/DC-498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lanesawyer (commented):
This fixes the issue, but I think the image check logic should live in `buildCompositor` instead so that we don't have two `ctx.putImageData(image, 0, 0)` lines floating around.

Not completely opposed to shipping it as-is though. The logic here isn't too complex. Just nervous that it'll grow as we polish up the viewer and then we'll be duplicating more and more instead of having a single function that does the proper drawings.

Jarbuckle:
> This fixes the issue, but I think the image check logic should live in `buildCompositor` instead so that we don't have two `ctx.putImageData(image, 0, 0)` lines floating around.
> 
> Not completely opposed to shipping it as-is though. The logic here isn't too complex. Just nervous that it'll grow as we polish up the viewer and then we'll be duplicating more and more instead of having a single function that does the proper drawings.

Jarbuckle:
I've made updates in the `bkp-client` PR first, @lanesawyer -- could you take a look over there? If it's good there, I'll make updates on this side.

lanesawyer (approved):
Looks good after merge!

PR #96, opened 2025-03-05
chore: Fix all but non-null assertion lints

lanesawyer:
# What

- Fixes all remaining lints besides the "non-null assertion" lints
- Leaves those 9 assertions for another time. We'll need to decide how to respond in each case if the item is undefined

# How

- `pnpm lint`
- Look at the problems
- Most of the time Biome suggested a solution and it worked perfectly!

# PR Checklist

### DIFF
--- a/packages/geometry/src/spatialIndexing/tree.ts
+++ b/packages/geometry/src/spatialIndexing/tree.ts
@@ -16,12 +16,12 @@ export function visitBFS<Tree>(
     while (frontier.length > 0) {
         const cur = frontier.shift()!;
         visitor(cur);
-        children(cur).forEach((c) => {
+        for (const c of children(cur)) {
             if (traversalPredicate?.(c) ?? true) {
                 // predicate?.(c) is true false or undefined. if its undefined, we coerce it to true with ??
                 // because we want to always traverse children if the predicate isn't given
                 frontier.push(c);
             }
-        });
+        }
     }
 }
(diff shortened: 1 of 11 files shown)


packages/geometry/src/spatialIndexing/tree.ts
```diff
@@ -16,12 +16,12 @@ export function visitBFS<Tree>(
     while (frontier.length > 0) {
         const cur = frontier.shift()!;
         visitor(cur);
-        children(cur).forEach((c) => {
+        for (const c of children(cur)) {
```
jessicamalloy, line 19: Are you replacing `forEach` with for loops to improve performance? Or is it a style preference?

packages/scatterbrain/src/abstract/render-server.ts
```diff
@@ -160,7 +160,7 @@ export class RenderServer {
     beginRendering<D, I>(renderFn: RenderFrameFn<D, I>, callback: ServerCallback<D, I>, client: Client) {
         if (this.regl) {
             const clientFrame = this.clients.get(client);
-            if (clientFrame && clientFrame.frame) {
+            if (clientFrame?.frame) {
```
jessicamalloy, line 163: 😌

jessicamalloy (approved):
Good job, there are lots of satisfying changes in this!

packages/geometry/src/spatialIndexing/tree.ts
lanesawyer: Yeah, there are a few reasons they have that rule, you can see more here! https://biomejs.dev/linter/rules/no-for-each/

PR #101, opened 2025-03-17
fix: Remove non-null assertions

lanesawyer:
# What

- Removes the remaining lint errors around non-null assertions

# How

- Look at each lint
- Remove the `!`
- Throw errors, continue working, or log warnings depending on how I best thought to address each. **THESE NEED REVIEW AND CONSENSUS** 

# PR Checklist


packages/scatterbrain/src/abstract/render-server.ts
```diff
@@ -47,7 +48,7 @@ type Client = HTMLCanvasElement;
 export class RenderServer {
     private canvas: OffscreenCanvas;
     private refreshRequested: boolean;
-    regl: REGL.Regl | null;
```
lanesawyer, line 50: This change seems drastic, but we set `regl` in the `constructor` so it should be fine. I could go further and remove a bunch of `if(!regl)` sort of checks too but kept it simple for now

PR #123, opened 2025-04-04
feature: Support for arbitrary color channels in OME-Zarr images [DC-530]

Jarbuckle:
# feature: Support for arbitrary color channels in OME-Zarr images [[DC-530](https://alleninstitute.atlassian.net/browse/DC-530)]

# What

**Ticket** - [DC-530](https://alleninstitute.atlassian.net/browse/DC-530)

Previously, the OME-Zarr renderer capabilities of Vis were hard-coded to only support RGB color channels in the source images. As there are known cases of images that use other colorspaces (such as CFP/YFP found in VS200 images), there was a need to implement support for arbitrary color channel rendering in Vis's `omezarr` package. This PR implements that feature.

# How

- Renamed the existing `buildTileRenderer` to `buildRGBTileRenderer`.
- Created a new `buildTileRenderer` which is given a specified number of channels, and then generates a shader based on that number of channels, additively building up the fragment color from the values and supplied RGB color mapping of each channel. (**NOTE:** I tried various cleaner approaches to implementing this, but due to issues with driver support/GLSL version/etc., I was unable to supply the shader with an array of channels, and instead had to specify the component elements of each channel as separate `uniform` parameters. This unfortunately also means that the `uniforms` are not typed as well as they could be... _sigh_ ).
- Updated the `SliceRenderer` system to use the new `buildTileRenderer`, altering types and infrastructure where needed to support arbitrary color channels.
- Updated the examples to use the new rendering mechanism.

# Testing

## Where
The example apps, particularly the OME-Zarr example but also the Layers example. For comparison, you may need to also start up the example apps for the current `main` branch.

## What
- Open the example app and compare outputs

## Expected
- Exactly the same output
- More thorough testing of behavior will be possible with the `bkp-client` update coming soon.

# PR Checklist


packages/omezarr/src/sliceview/slice-renderer.ts
```diff
@@ -91,46 +113,53 @@ export function buildOmeZarrSliceRenderer(
...
         };
     }
-    const cmd = buildTileRenderer(regl);
+    const cmd = buildTileRenderer(regl, numChannels);
     return {
         cacheKey: (item, requestKey, dataset, settings) => {
-            const col = requestKey as keyof RenderSettings['gamut'];
-            const index = settings.gamut[col]?.index ?? 0;
-            return `${dataset.url}_${JSON.stringify(item)}_ch=${index.toFixed(0)}`;
+            return `${dataset.url}_${JSON.stringify(item)}_ch=${requestKey}`;
```
lanesawyer, line 119: Maybe I'll see it elsewhere as I go through the review, but I'm not fully confident in this. Is the `requestKey` the channel in this new world of fully supported color channels?

packages/omezarr/src/sliceview/slice-renderer.ts
```diff
@@ -91,46 +113,53 @@ export function buildOmeZarrSliceRenderer(
...
         },
         fetchItemContent: (item, dataset, settings, signal) => {
-            return {
-                R: () => decoder(dataset, toZarrRequest(item, settings.gamut.R.index), item.level).then(sliceAsTexture),
-                G: () => decoder(dataset, toZarrRequest(item, settings.gamut.G.index), item.level).then(sliceAsTexture),
-                B: () => decoder(dataset, toZarrRequest(item, settings.gamut.B.index), item.level).then(sliceAsTexture),
-            };
+            const keys = keysOf(settings.channels);
+            const result = keys
+                .map((key) => ({
```
lanesawyer, line 129: I think this can be accomplished with a single `reduce` instead of a `map` and then `reduce`

packages/omezarr/src/sliceview/tile-renderer.ts
```diff
@@ -17,14 +17,28 @@ type Props = {
...
     B: REGL.Texture2D;
 };
+
+type Channel = {
+    tex: REGL.Texture2D;
+    gamut: vec2;
+    color: vec3;
+};
+
+type GenericProps = {
```
lanesawyer, line 27: Kinda depend on whether you think we'll have the RBG Tile Renderer long-term, but I think these props could use some renames and combining.

Along these lines?

```
type SharedProps = {
    target: Framebuffer2D | null;
    tile: vec4; // [minx,miny,maxx,maxy] representing the bounding box of the tile we're rendering
    view: vec4; // [minx,miny,maxx,maxy] representing the camera in the same space as the tile's bounding box
}

type RgbTileRendererProps = SharedProps & {
    Rgamut: vec2; // [min,max] RedOut = RedChannelValue-Rgamut.min/(Rgamut.max-Rgamut.min)
    Ggamut: vec2; // [min,max] GreenOut = GreenChannelValue-Ggamut.min/(Ggamut.max-Ggamut.min)
    Bgamut: vec2; // [min,max] BlueOut = BlueChannelValue-Bgamut.min/(Bgamut.max-Bgamut.min)
    R: REGL.Texture2D;
    G: REGL.Texture2D;
    B: REGL.Texture2D;
};

type Channel = {
    tex: REGL.Texture2D;
    gamut: vec2;
    color: vec3;
};

type TileRendererProps = SharedProps & {
    channels: Channel[];
};
```

lanesawyer (commented):
Existing examples work great! A few minor comments for you.

Another nice thing would be to get one of the new OME-Zarr files that this enables rendering for and add it as an option in the example. We don't need to get so fancy with a dropdown or something, but even just a bulleted list of a few URLs below the viewer (for now) would be a nice way to allow quick testing of the different types

packages/omezarr/src/sliceview/slice-renderer.ts
```diff
@@ -62,22 +76,30 @@ function toZarrRequest(tile: VoxelTile, channel: number): ZarrRequest {
...
+
+export type OmeZarrSliceRendererOptions = {
+    numChannels?: number;
+};
+const DEFAULT_NUM_CHANNELS = 3;
+
 export function buildOmeZarrSliceRenderer(
     regl: REGL.Regl,
     decoder: Decoder,
+    options?: OmeZarrSliceRendererOptions | undefined,
```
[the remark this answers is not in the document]
Jarbuckle: To be honest, I'm not quite sure how TypeScript handles situations like this. An optional argument is, by definition, potentially `undefined`, but I feel like there have been instances where I left the `| undefined` off and TypeScript/some linter shouted at me. Consequently, I've just been tacking `| undefined` on there as a safety/explicitness measure.

Strictly speaking, of course, there's technically no difference in the underlying JS between simply leaving off an optional argument and passing in `undefined` to that argument, at least as I understand it. If the `?` is just a shorthand for appending `| undefined` to a type, then I'm fine with leaving it off.

packages/omezarr/src/sliceview/tile-renderer.ts
Jarbuckle: Yep, good call! I did a renaming towards the end and missed these 😅

packages/omezarr/src/sliceview/slice-renderer.ts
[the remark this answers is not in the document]
lanesawyer: LLMs told me that there really isn't a difference, so either way is fine by me! I slightly prefer the less verbose one though 😁

packages/omezarr/src/sliceview/slice-renderer.ts
Jarbuckle: Looking it over, I think I'm inclined to instead use the `for...in` loop for this. It's recommended by Biome, anyway 🤷🏽

packages/omezarr/src/sliceview/tile-renderer.ts
Jarbuckle: Done!

packages/omezarr/src/sliceview/slice-renderer.ts
Jarbuckle: Updated version pushed!

packages/omezarr/src/sliceview/slice-renderer.ts
Jarbuckle: I walked through the relationship between the `cacheKey` and `fetchItemContent` functions, and confirmed that there should be a 1:1 relationship between the keys used in each. However, for the sake of clarity, I added a value check into `cacheKey` to make sure that only valid keys are accepted.

Jarbuckle:
I'll add the extra example files on Monday 🙂

packages/omezarr/src/sliceview/slice-renderer.ts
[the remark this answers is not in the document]
Jarbuckle: Alrighty, well, there was at least one place where Typescript insisted on me adding `| undefined`, but I've tried to leave it out where I can.

examples/src/omezarr/omezarr-demo.tsx
lanesawyer, line 1: Love this! Much more than I was expecting 😁

packages/core/src/colors.ts
lanesawyer, line 1: It would be nice to have tests for these, but let's skip it for now. It looks to be working, and since we have the test coverage installed now we'll easily remember that this could use a little bit of unit testing later!

packages/core/src/resources.ts
lanesawyer, line 1: Also another piece that could use some basic unit tests, but we'll save it for another day. These look correct and work!

lanesawyer (approved):
Looks great, saw the aspect ratio issue you mentioned, we'll keep an eye on it but if things look find in the `bkp-client` we'll definitely save that for another day.

packages/core/src/colors.ts
Jarbuckle: Also, there's Scatterbrain-related color stuff in `bkp-client` right now which overlaps some with this, some merging that work into this module would be a nice thing to do around the same time, I'd say 🙂

PR #126, opened 2025-04-15
Noah/webworker decoders

froyo-np:
# WIP Steps

- [x] sometimes never finishes the frame... figure out why?
- [x] update the viewer to use a preview buffer during zooms
- [x] prefech the lowest LOD for all slides 
- [x] update version
# What
- [x] use a stashed image to hide in-progress loading behind a stand-in version
- [x] use a pool of web-workers to decouple ome-zarr fetching and decoding from the ui-thread
- [x] use a lower threshold for in-flight tasks, which should result in less stress on the S3 bucket serving our data.
 

# Screenshots

# PR Checklist

### DIFF
--- a/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
+++ b/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
@@ -2,7 +2,6 @@ import { loadSlice, OmeZarrMetadata } from '@alleninstitute/vis-omezarr';
 import { logger } from '@alleninstitute/vis-core';
 import type { Chunk, Float32 } from 'zarrita';
 import { isCancellationRequest, isSliceRequest } from './types';
-
 // a web-worker which fetches slices of data, decodes them, and returns the result as a flat float32 array, using transferables
 
 const ctx = self;
--- a/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
+++ b/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
@@ -1,30 +1,32 @@
-import { loadSlice, OmeZarrMetadata } from '@alleninstitute/vis-omezarr';
+import { loadSlice, makeOmeZarrSliceLoaderWorker, OmeZarrMetadata } from '@alleninstitute/vis-omezarr';
 import { logger } from '@alleninstitute/vis-core';
 import type { Chunk, Float32 } from 'zarrita';
 import { isCancellationRequest, isSliceRequest } from './types';
 // a web-worker which fetches slices of data, decodes them, and returns the result as a flat float32 array, using transferables
 
 const ctx = self;
-const cancelers: Record<string, AbortController> = {}
-ctx.onmessage = (msg: MessageEvent<unknown>) => {
-    const { data } = msg;
-    try {
-        if (isSliceRequest(data)) {
-            const { metadata: dehydratedMetadata, req, level, id } = data;
-            const abort = new AbortController();
-            cancelers[id] = abort;
-            OmeZarrMetadata.rehydrate(dehydratedMetadata).then((metadata) => {
-                loadSlice(metadata, req, level, abort.signal).then((result: { shape: number[]; buffer: Chunk<Float32> }) => {
-                    const { shape, buffer } = result;
-                    const data = new Float32Array(buffer.data);
-                    ctx.postMessage({ type: 'slice', id, shape, data }, { transfer: [data.buffer] });
-                });
-            });
-        } else if (isCancellationRequest(data)) {
-            const { id } = data;
-            cancelers[id]?.abort("cancelled");
-        }
-    } catch (err) {
-        logger.error('OME-Zarr fetch onmessage error', err);
-    }
-};
+// const cancelers: Record<string, AbortController> = {}
+
+// ctx.onmessage = (msg: MessageEvent<unknown>) => {
+//     const { data } = msg;
+//     try {
+//         if (isSliceRequest(data)) {
+//             const { metadata: dehydratedMetadata, req, level, id } = data;
+//             const abort = new AbortController();
+//             cancelers[id] = abort;
+//             OmeZarrMetadata.rehydrate(dehydratedMetadata).then((metadata) => {
+//                 loadSlice(metadata, req, level, abort.signal).then((result: { shape: number[]; buffer: Chunk<Float32> }) => {
+//                     const { shape, buffer } = result;
+//                     const data = new Float32Array(buffer.data);
+//                     ctx.postMessage({ type: 'slice', id, shape, data }, { transfer: [data.buffer] });
+//                 });
+//             });
+//         } else if (isCancellationRequest(data)) {
+//             const { id } = data;
+//             cancelers[id]?.abort("cancelled");
+//         }
+//     } catch (err) {
+//         logger.error('OME-Zarr fetch onmessage error', err);
+//     }
+// };
+makeOmeZarrSliceLoaderWorker(ctx);
--- a/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
+++ b/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
@@ -1,24 +1,29 @@
 import { loadSlice, OmeZarrMetadata } from '@alleninstitute/vis-omezarr';
 import { logger } from '@alleninstitute/vis-core';
 import type { Chunk, Float32 } from 'zarrita';
-import { isSliceRequest } from './types';
+import { isCancellationRequest, isSliceRequest } from './types';
 
 // a web-worker which fetches slices of data, decodes them, and returns the result as a flat float32 array, using transferables
 
 const ctx = self;
-
+const cancelers: Record<string, AbortController> = {}
 ctx.onmessage = (msg: MessageEvent<unknown>) => {
     const { data } = msg;
     try {
         if (isSliceRequest(data)) {
             const { metadata: dehydratedMetadata, req, level, id } = data;
+            const what = new AbortController();
+            cancelers[id] = what;
             OmeZarrMetadata.rehydrate(dehydratedMetadata).then((metadata) => {
-                loadSlice(metadata, req, level).then((result: { shape: number[]; buffer: Chunk<Float32> }) => {
+                loadSlice(metadata, req, level, what.signal).then((result: { shape: number[]; buffer: Chunk<Float32> }) => {
                     const { shape, buffer } = result;
                     const data = new Float32Array(buffer.data);
                     ctx.postMessage({ type: 'slice', id, shape, data }, { transfer: [data.buffer] });
                 });
             });
+        } else if (isCancellationRequest(data)) {
+            const { id } = data;
+            cancelers[id]?.abort("cancelled");
         }
     } catch (err) {
         logger.error('OME-Zarr fetch onmessage error', err);
--- a/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
+++ b/examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
@@ -1,32 +1,6 @@
-import { loadSlice, makeOmeZarrSliceLoaderWorker, OmeZarrMetadata } from '@alleninstitute/vis-omezarr';
-import { logger } from '@alleninstitute/vis-core';
-import type { Chunk, Float32 } from 'zarrita';
-import { isCancellationRequest, isSliceRequest } from './types';
+import { makeOmeZarrSliceLoaderWorker } from '@alleninstitute/vis-omezarr';
 // a web-worker which fetches slices of data, decodes them, and returns the result as a flat float32 array, using transferables
 
 const ctx = self;
-// const cancelers: Record<string, AbortController> = {}
 
-// ctx.onmessage = (msg: MessageEvent<unknown>) => {
-//     const { data } = msg;
-//     try {
-//         if (isSliceRequest(data)) {
-//             const { metadata: dehydratedMetadata, req, level, id } = data;
-//             const abort = new AbortController();
-//             cancelers[id] = abort;
-//             OmeZarrMetadata.rehydrate(dehydratedMetadata).then((metadata) => {
-//                 loadSlice(metadata, req, level, abort.signal).then((result: { shape: number[]; buffer: Chunk<Float32> }) => {
-//                     const { shape, buffer } = result;
-//                     const data = new Float32Array(buffer.data);
-//                     ctx.postMessage({ type: 'slice', id, shape, data }, { transfer: [data.buffer] });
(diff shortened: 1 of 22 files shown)


examples/src/omezarr/omezarr-viewer.tsx
```diff
@@ -1,114 +1,117 @@
...
+            case 'progress':
+              // wanna see the tiles as they arrive?
+              e.server.copyToClient(compose);
+              break;
+            case 'finished': {
+              e.server.copyToClient(compose);
+              break;
+            }
+            case 'cancelled':
+              console.log('frame cancelled!');
```
froyo-np, line 92: TODO remove me - this was for my benefit

examples/src/common/loaders/ome-zarr/sliceWorkerPool.ts
```diff
@@ -72,7 +74,13 @@ export class SliceWorkerPool {
                 req,
                 level,
             };
-            this.workers[this.which].postMessage(message);
+            if (signal) {
+                signal.onabort = (ev) => {
+                    this.workers[myWorker].postMessage({ type: 'cancel', id: reqId })
+                    this.promises[reqId]?.reject("cancelled")
```
froyo-np, line 80: hmmm - todo: "delete this.promises[reqId]" ?

examples/src/common/loaders/ome-zarr/fetchSlice.worker.ts
```diff
@@ -1,24 +1,28 @@
...
 
 const ctx = self;
-
+const cancelers: Record<string, AbortController> = {}
 ctx.onmessage = (msg: MessageEvent<unknown>) => {
     const { data } = msg;
     try {
         if (isSliceRequest(data)) {
             const { metadata: dehydratedMetadata, req, level, id } = data;
+            const what = new AbortController();
```
froyo-np, line 14:
```suggestion
            const justKiddingDontFetch = new AbortController();
```

packages/omezarr/src/zarr/loading.ts
```diff
@@ -314,7 +314,7 @@ export async function loadSlice(metadata: OmeZarrMetadata, r: ZarrRequest, level
         throw new VisZarrDataError(message);
     }
     const { raw } = await loadZarrArrayFileFromStore(store, arr.path, 2, false);
-    const result = await zarr.get(raw, buildQuery(r, axes, level.shape));
+    const result = await zarr.get(raw, buildQuery(r, axes, level.shape), { opts: { signal: signal ?? null } });
```
froyo-np, line 317: zarrita is delightful and allows us to pass an abort signal! nice!

lanesawyer:
@froyo-np I have a fix for the web worker error in my PR to get the deploy site working again (#127)

froyo-np:
> @froyo-np I have a fix for the web worker error in my PR to get the deploy site working again (#129)

ha I just independantly found the same thing! nice, and would this let us move the worker into the packages ?

froyo-np:
> > @froyo-np I have a fix for the web worker error in my PR to get the deploy site working again (#129)
> 
> ha I just independantly found the same thing! nice, and would this let us move the worker into the packages ?

no - I forgot we use parcel for that

examples/src/omezarr/omezarr-viewer.tsx
```diff
@@ -44,30 +50,96 @@ export function OmezarrViewer({
...
+    const imgRenderer = useRef<ReturnType<typeof buildImageRenderer>>();
+    const stash = useRef<StashedView>();
 
     // setup renderer and delete it when component goes away
     useEffect(() => {
         const c = canvas?.current;
         if (server?.regl && omezarr) {
             const numChannels = omezarr.colorChannels.length || 3;
-            renderer.current = buildAsyncOmezarrRenderer(server.regl, defaultDecoder, { numChannels });
+            renderer.current = buildAsyncOmezarrRenderer(server.regl, multithreadedDecoder, {
```
froyo-np, line 61: here's a big piece of the performance upgrade - using the web-worker based multithreadedDecoder.

examples/src/omezarr/omezarr-viewer.tsx
```diff
@@ -44,30 +50,96 @@ export function OmezarrViewer({
...
+                numChannels,
+                queueOptions: { maximumInflightAsyncTasks: 1 },
```
froyo-np, line 63: I also exposed the queue control values as options when creating the renderer - I think the default value here was letting a great many requests hammer the S3 server, leading to slowness when rapidly canceling frames.

examples/src/omezarr/omezarr-viewer.tsx
```diff
@@ -44,30 +50,96 @@ export function OmezarrViewer({
...
         return () => {
             if (c) {
                 server?.destroyClient(c);
             }
         };
     }, [server, omezarr]);
+    useEffect(() => {
+        // set up the stash:
+        if (server?.regl) {
+            stash.current = {
```
froyo-np, line 76: here's the smoke-and-mirrors performance improvement: we keep a copy of the rendered screen in "stash".

examples/src/omezarr/omezarr-viewer.tsx
```diff
@@ -79,15 +151,36 @@ export function OmezarrViewer({
                                 color: [0, 0, 0, 0],
                                 depth: 1,
                             });
+                            lowResPreview(e.target, server.cache, (_e) => {})?.cancelFrame(
```
froyo-np, line 154: here, whenever we begin a frame, we first:
1. request a low-resolution preview frame be rendered to our target. By immediately cancelling, we ensure that we only render things that are already in the cache. If they're not in the cache, they'll still be requested, so maybe they'll be present in the future.
2. on line 157, we render the stashed image ABOVE the low-res preview. This hides the "pixelated flickering" that would otherwise be caused by the low-res preview.
3. finally, we proceed as normal, rendering the actual frame we want above both the low-res preview and the stashed copy.

packages/omezarr/src/sliceview/slice-renderer.ts
```diff
@@ -144,24 +152,34 @@ export function buildOmeZarrSliceRenderer(
...
+            // determine the depth at which to render - the goal is that lower resolution tiles should be rendered further back
+            const wholeLayerResolution = planeSizeInVoxels(
+                settings.plane,
+                dataset.attrs.multiscales[0].axes,
+                item.level,
+            );
+            // this number can be very very large - the goal is to have a number from [0:1]...
+            // ideally, we'd go figure out the size of the same plane of the largest layer, but that is annoying right now
+            const divisor = 10_000_000;
+            const depth = Vec2.scale(wholeLayerResolution ?? [0, 0], 1 / divisor);
```
froyo-np, line 177: this change renders tiles at a Z-depth that is linearly related to their resolution. My plan here was to stop fiddling around with ordered rendering and just let the depth buffer sort it all out. Theres weird annoying math here - item.level used to be an integer telling us which layer its on, but now its a whole shapedDataset.

This isnt a bad thing, and its harmless, but not really used right now, so maybe I should revert it? thoughts?

packages/omezarr/src/sliceview/tile-renderer.ts
```diff
@@ -6,8 +6,30 @@ import type { vec2, vec3, vec4 } from '@alleninstitute/vis-geometry';
 import type REGL from 'regl';
 import type { Framebuffer2D } from 'regl';
 
+const vert = /*glsl*/ `
```
froyo-np, line 9: this vertex shader was duplicated between the two commands in this file, so I just dry'd it up a bit

packages/core/src/abstract/async-frame.ts
```diff
@@ -183,6 +183,16 @@ export function beginFrame<
         },
     };
 }
+type QueueOptions = {
+    queueProcessingIntervalMS?: number;
+    maximumInflightAsyncTasks?: number;
+    queueTimeBudgetMS?: number;
+};
+const defaultQueueOptions = {
```
Jarbuckle, line 191: Even though it's practically the same thing, I would suggest adding an explicit type of `defaultQueueOptions` so that intellisense and such will know to display it as a `QueueOptions` value rather than listing out its properties -- unless of course you specifically want those property values to be viewable in tooling?

packages/omezarr/src/sliceview/slice-renderer.ts
```diff
@@ -144,24 +152,34 @@ export function buildOmeZarrSliceRenderer(
...
-
+            // determine the depth at which to render - the goal is that lower resolution tiles should be rendered further back
+            const wholeLayerResolution = planeSizeInVoxels(
+                settings.plane,
+                dataset.attrs.multiscales[0].axes,
+                item.level,
+            );
+            // this number can be very very large - the goal is to have a number from [0:1]...
+            // ideally, we'd go figure out the size of the same plane of the largest layer, but that is annoying right now
+            const divisor = 10_000_000;
```
Jarbuckle, line 176: Is this more feasible to calculate now, with the way the metadata's loaded in, or is it still awkward?

packages/omezarr/src/sliceview/tile-renderer.ts
```diff
@@ -102,6 +107,7 @@ export function buildRGBTileRenderer(regl: REGL.Regl) {
         uniforms: {
             tile: regl.prop<RGBTileRenderProps, 'tile'>('tile'),
             view: regl.prop<RGBTileRenderProps, 'view'>('view'),
+            depth: regl.prop<RGBTileRenderProps, 'depth'>('depth'),
```
Jarbuckle, line 110: Just for my own edification, I'm curious about these depth buffer additions; how's that being used in this scenario?

packages/omezarr/src/sliceview/worker-loader.ts
```diff
@@ -0,0 +1,60 @@
...
+}
+/**
+ * a helper function to initialize a message handler on a webworker,
+ * which responds to requests for omezarr slices:
+ * messages must be of type MessageEvent<ZarrSliceRequest|CancelRequest>
+ * @see ZarrSliceRequest
+ * @see CancelRequest
+ * @param ctx the "global this" aka self object on a webworker context.
+ */
+export function makeOmeZarrSliceLoaderWorker(ctx: typeof self) {
```
Jarbuckle, line 33: `typeof self` seems like it might be a bit unstable outside of the context of a web worker. Would it make sense to switch this to the `WorkerGlobalScope` type, if it exists? https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/self

packages/omezarr/src/sliceview/worker-loader.ts
```diff
@@ -0,0 +1,60 @@
...
+                            const { shape, buffer } = result;
+                            const data = new Float32Array(buffer.data);
+                            ctx.postMessage({ type: 'slice', id, shape, data }, { transfer: [data.buffer] });
+                        },
+                    );
+                });
+            } else if (isCancellationRequest(data)) {
+                const { id } = data;
+                cancelers[id]?.abort('cancelled');
+            }
```
Jarbuckle, line 55: Wondering if we should handle the `else` scenario here -- maybe just treat it like a cancellation? Not sure if web workers are able to actually output any logs 🤔

Jarbuckle (commented):
Code review done, got a few thoughts/suggestions for you to take a look at!

Jarbuckle:
Running the example code, some observations:
- It seems to me that the overall loading process is significantly slower than it was when I lasted looked at this branch; it's almost like the webworkers aren't kicking in, or are running slower than they were?
- I'm not sure the cancellation of in-progress webworkers is functioning (possibly related to the issue above), as trying to skip past a bunch of slides results in all of them loading anyway, or at least that's what seems to be happening 🤔

examples/src/data-renderers/versa-renderer.ts
```diff
@@ -219,7 +219,7 @@ export function cacheKeyFactory(col: string, item: VoxelTile, settings: VoxelSli
 }
 
 function reqSlice(dataset: OmeZarrMetadata, req: ZarrRequest, layerIndex: number) {
-    return getSlicePool().requestSlice(dataset, req, layerIndex);
+    return getSlicePool().requestSlice(dataset, req, dataset.getAllShapedDatasets(0)[layerIndex]);
```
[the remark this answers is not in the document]
froyo-np: ah nice

packages/core/src/abstract/async-frame.ts
froyo-np: I'm fine listing out its properties, although its not exported... so I dont think anyone will have any visibility of it unless they're reading this file

packages/omezarr/src/sliceview/slice-renderer.ts
froyo-np: hard to say - what I want is a number between 0 and 1 in which lower-res layers are near 1 and the highest resolution layer is 0. it was "easy" when level was the index of the layer (although I was assuming that the layers in the omezarr are in order...). as it is now, it would be better go find the largest layer explicitly, rather than use a big number as a guess.

packages/omezarr/src/sliceview/tile-renderer.ts
froyo-np: previously, all tiles draw at the same depth, with depth testing off, which means that tiles that overlap are rendered "last write wins" style. My plan here was to use the size of the layers in the pyramid to make lower resolution tiles always render behind higher resolution layers, regardless of the order in which they're dispatched ot the gpu.

packages/omezarr/src/sliceview/worker-loader.ts
froyo-np: how do you mean unstable? As far as I can tell, self is what I want?
https://developer.mozilla.org/en-US/docs/Web/API/Window/self

packages/omezarr/src/sliceview/worker-loader.ts
froyo-np: ah excellent call, will do - its pretty dang easy to mess up what messages you pass to a worker!

packages/omezarr/src/sliceview/slice-renderer.ts
froyo-np: I've removed this

Jarbuckle:
Took another look at the example app, and noticing a few new things:
- It seems that the renderer frequently gets "stuck", only rendering very low-res images after zooming in; it only infrequently loads the higher-resolution data. 😕 Moving the camera around the way you showed me yesterday doesn't seem to help with this one. 🙁
- Are we going to go with the `maximumInflightAsyncTasks` of `1`, or switch to `2` like we chatted about yesterday? I'm currently seeing about the same slower speed of load times as I was seeing yesterday when it was set to `1`.

Jarbuckle (commented):
Just one thought on the code, related to the loading issue mentioned above!

PR #158, opened 2025-05-21
Some quick minor fixes to the examples in starlight

froyo-np:
# What
* make everything wider
* prevent scrolling when trying to zoom on a visualization
* fix a crash when changing the plane (e.g xy ->yz) on an ome-zarr slice view in the layers demo.

# How

# Screenshots

# PR Checklist


site/custom.css
```diff
@@ -0,0 +1,3 @@
+:root {
+    --sl-content-width: 100rem;
```
froyo-np, line 2: perhaps the most controversial change!

site/src/examples/omezarr/omezarr-demo.tsx
```diff
@@ -46,7 +45,7 @@ const demoOptions: DemoOption[] = [
     },
 ];
 
-const screenSize: vec2 = [500, 500];
+const screenSize: vec2 = [900, 900];
```
TheMooseman, line 48: I think this might be too tall, I can't initially see the JSON below the image, which I think we'd want people to see so that they can understand a bit more about what it's doing

TheMooseman (approved):
LGTM, just need to fix the formatting in the layers demo file. I think the omezarr previewer should be smaller, but I'm not gonna block the PR on that

site/custom.css
lanesawyer: Yeah, this ends up nuking the helpful navigation for the markdown docs:
Existing
image

New
image

I love the idea of making the examples bigger, let's just scope this change to only those ones. I think we can accomplish that with a custom layout for those pages

lanesawyer (changes requested):
I like the changes except for what happens to the docs layout! Let's scope the wider content area to just the demos

PR #162, opened 2025-05-28
feat: DZI fetch function

lanesawyer:
# What

- A `fetch` helper for DZI image code to simplify getting started with the `vis-dzi` package

# How

- Took `bkp-client` fetch code
- Spruced it up a bit with our `vis` `logger`,
- Made a helper function in `dzi/loader.ts` that is exported

# PR Checklist

### DIFF
--- a/packages/dzi/src/loader.ts
+++ b/packages/dzi/src/loader.ts
@@ -1,6 +1,10 @@
+import { logger } from '@alleninstitute/vis-core';
 import { Box2D, type Interval, Vec2, type box2D, type vec2 } from '@alleninstitute/vis-geometry';
 
 type DziTilesRoot = `${string}_files/`;
+type DziFormat = 'jpeg' | 'png' | 'jpg' | 'JPG' | 'PNG';
+const isDziFormat = (format: string): format is DziFormat => ['jpeg', 'png', 'jpg', 'JPG', 'PNG'].includes(format);
+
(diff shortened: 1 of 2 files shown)


froyo-np (approved):
looks great to me!

packages/dzi/src/loader.ts
```diff
@@ -1,14 +1,18 @@
+import { logger } from '@alleninstitute/vis-core';
 import { Box2D, type Interval, Vec2, type box2D, type vec2 } from '@alleninstitute/vis-geometry';
 
 type DziTilesRoot = `${string}_files/`;
+type DziFormat = 'jpeg' | 'png' | 'jpg' | 'JPG' | 'PNG';
+const isDziFormat = (format: string): format is DziFormat => ['jpeg', 'png', 'jpg', 'JPG', 'PNG'].includes(format);
```
froyo-np, line 6: a minor suggestion - avoid repeating the list of file type strings like so:
```
const formats = ['jpeg', 'png' ... etc] as const
type DziFormat = ElementType<typeof formats>
```

lanesawyer: Hm, it's complaining about the `.includes` being a string and it only wants the `formats` with some of those changes. I can do more TS wizardry to make it work, but I think I'll just leave the duplicated string version as-is.

If/when we need that list in more places, we'll revisit some fancy typing!

PR #171, opened 2025-06-30
A priority cache with a (better?) api

froyo-np:
# WIP Steps

_This section is optional if the PR is fully ready for review_

-   Create the PR as a "draft" PR
-   Add [WIP] to the title
-   Use the template below to ask for specific feedback
-   [ ] @name: Could you look at \_\_\_? I'm trying to do... (etc.)

# What

Replace this with 1-2 line Description of what feature, bugfix, chore does the PR contain the code for.

# How

# Screenshots

# PR Checklist


packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,143 @@
...
+function mapFields<R extends Record<string, any>, Result>(r: R, fn: (v: ObjectValue<R>) => Result): { [k in keyof R]: Result } {
+    return entries(r).reduce((acc, [k, v]) => ({ ...acc, [k]: fn(v) }), {} as { [k in keyof R]: Result })
+}
+
+type Client = {
+    low: Set<string>
+    high: Set<string>
+    notify: undefined | ((cacheKey: string) => void)
+}
+export class FancySharedCache {
```
froyo-np, line 38: here is the shared cache - I should rename it... maybe just SharedPriorityCache?

packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,143 @@
...
+            const keys = spec.cacheKeys(item) as Record<string, string>
+            const fetchers = spec.fetch(item)
+            return Object.keys(keys).map((sk) => ({ cacheKey: keys[sk], fetcher: fetchers[sk] }))
+        }
+
+        const priorityDelta = (key: string, pri: 0 | 1 | 2, low: Set<string>, high: Set<string>) => {
+            return pri - (high.has(key) ? 2 : (low.has(key) ? 1 : 0))
+        }
+        const setPriorities = (low: Set<Key>, high: Set<Key>) => {
+            const newLow = new Set<string>()
```
froyo-np, line 61: here is the meat of the new approach - every time a client updates its priorities, the system has to re-balance the heap of all priorities. This also has a bug due to how Set() deals with equality, and thus membership

todo: rewrite and test more

packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,143 @@
+import * as lo from 'lodash' // vite/astro went bonkers - not sure why this helps, core has always had lodash in it...
+import { PriorityCache, type Store, type Resource } from "./priority-cache";
+const { uniqueId } = lo
+// goal: we want clients of the cache to experience a type-safe interface -
+// they expect that the things coming out of the cache are the type they expect (what they put in it)
+// this is not strictly true, as the cache is shared, and other clients may use different types
+// also also - there will not be a 1:1 relation between items and
+type CacheInterface<Key, Value extends Record<string, Resource>> = {
```
froyo-np, line 8: todo: rename generic parameter Key here - like our current cache, we let the users of the cache deal in objects which can require multiple, distinct entries in the cache itself.

site/src/examples/omezarr/omezarr-via-priority-cache.tsx
```diff
@@ -0,0 +1,291 @@
...
+    isValue: (v): v is Record<string, Tex> =>
+      renderer.isPrepared(
+        mapValues(v, (tx: Resource | undefined) => (tx && tx instanceof Tex ? tx.texture : undefined))
+      ),
+    onDataArrived: onData,
+  });
+  return {
+    render: (target: REGL.Framebuffer2D | null, dataset: OmeZarrMetadata, settings: RenderSettings) => {
+      const items = renderer.getVisibleItems(dataset, settings);
+      const baselayer = renderer.getVisibleItems(dataset, {
```
froyo-np, line 76: 2. get the lowest resolution data for the same view

site/src/examples/omezarr/omezarr-via-priority-cache.tsx
```diff
@@ -0,0 +1,291 @@
...
+        ...settings,
+        camera: { ...settings.camera, screenSize: [1, 1] },
+      });
+      regl.clear({ framebuffer: target, color: [0, 0, 0, 1], depth: 1 });
+      client.setPriorities(
```
froyo-np, line 81: mark each set as low or high priority

site/src/examples/omezarr/omezarr-via-priority-cache.tsx
```diff
@@ -0,0 +1,291 @@
...
+        new Set(items.map((tile) => ({ tile, dataset, settings }))),
+        new Set(baselayer.map((tile) => ({ tile, dataset, settings })))
+      );
+      for (const tile of [...baselayer, ...items]) {
```
froyo-np, line 85: draw everything, in order, if we have it

lanesawyer: THIS is where it clicked for me.

With this sort of flexibility we should be able to get rid of the current major pixelation on zoom because we can always make sure to draw the existing zoomed out layer first while the others are loading, correct?

packages/core/src/dataset-cache.ts
```diff
@@ -225,18 +226,20 @@ export class AsyncDataCache<SemanticKey extends RecordKey, CacheKey extends Reco
                 return resolvedCacheData;
             }
         } else {
-            promise = getter();
+            const abort = new AbortController();
```
lanesawyer, line 229: I don't think AbortControllers are all that expensive, but this will be called often.

Should be fine, but if they are expensive, that could be an issue!

packages/core/src/shared-priority-cache/readme.md
lanesawyer, line 1: Good arguments, set the stage well for why this change makes sense so the PR review has been much easier. Thanks!

I think this could be moved into the `/vis/packages/core/` documentation site with a couple small tweaks and be great `vis` library developer documentation.

site/src/examples/common/react/priority-cache-provider.tsx
lanesawyer, line 1: This duplicates the existing `render-server-provder.tsx`, so that's a good sign that this proposal doesn't result in sweeping breaking API changes 😁

lanesawyer (commented):
Seeing no major issues with the approach taken here! I like it and think it will make our life easier going forward as we do more complex rendering approaches.

site/src/examples/omezarr/omezarr-via-priority-cache.tsx
froyo-np: yup - flexibility is exactly the goal - you're still stuck if you dont have what you want in the cache, but other than that you can do whatever you want - you could draw dots only if you have the pixels that go with them, you could draw some, clear the depth buffer, go get coffee, then come back and render the dots - whatever you can think of!

packages/core/src/dataset-cache.ts
froyo-np: they have not seemed to be expensive in my testing so far - this change does use a lot more of them though - the old system could only cancel on a per-item basis (one Item -> many cache entries), but for the fine grained sharing and prioritization, using an abort signal per cache entry made more sense

froyo-np: also also - even though we supported aborting fetches in the previous cache, it often wasnt connected, due to its poor behavior. You'd cancel a frame, that would abort a fetch for item X, and then the very next frame would fire up a fetch for item X :(  ... that should not happen anymore!

site/src/examples/omezarr/omezarr-via-priority-cache.tsx
froyo-np: I think my current plan (for that pixelation issue) would be to keep a list of tiles drawn recently - and use them to fill in holes while we wait for new stuff to load. In addition, this makes it easy to pre-fetch things - we could pre-fetch slices N-1 and N+1 simply by listing them as low-priority (and then simply not drawing them!)

packages/core/src/shared-priority-cache/priority-cache.ts
```diff
@@ -0,0 +1,157 @@
...
+    // add {key:item} to the cache - return false (and fail) if the key is already present
+    // may evict items to make room
+    // return true on success
+    put(key: CacheKey, item: Resource): boolean {
+        if (this.store.has(key)) {
+            return false;
+        }
+        const size = item.sizeInBytes?.() ?? 1;
+        if (this.used + size > this.limit) {
+            this.evictUntil(Math.max(0, this.limit - size));
```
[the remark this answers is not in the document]
froyo-np: evicting more than needed - I see what you mean, but it tends to work out as the equivalent of lowering the cache limit, which is confusing right ? if the limit is the limit, its better to just use the limit :)

as far as the cost of eviction - in terms of overhead, its very low. in terms of thrashyness, who's to say that thing you aggressively evicted wont be needed the very next frame? caches with smaller limits will churn more than caches with higher limits (the risk being of course that you actually do run out of memory)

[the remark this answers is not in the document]
froyo-np: think of it this way also - yes, we would like put() to happen frequently (that means data is arriving, and yes that means going through the eviction logic repeatedly if the cache is nearly full). The secret we know though, is that puts only happen at the end of a fetch(s3://huge-pile-of-data.zarr)! that means they're actually super duper rare. even though the upper levels of the system know that 50 puts are on the way, each one takes a lifetime. Wouldnt it be better to be able to have data (low priority data, granted) in the cache to serve up while we wait for each of those 50 `fetch().then(put)` s? If we delete 50 entries from the cache up front, when we know we want to put 50 entries in, theres a large period of time where cache misses are more likely.

packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,148 @@
...
+            (ck) => this.importance[ck] ?? 0,
+            limitInBytes,
+            max_concurrent_fetches,
+            (ck) => this.onCacheEntryArrived(ck),
+        );
+    }
+    registerClient<Key, Value extends Record<string, Resource>>(
+        spec: ClientSpec<Key, Value>,
+    ): CacheInterface<Key, Value> {
+        const id = uniqueId('client');
```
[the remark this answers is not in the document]
froyo-np: yeah - weird back and forth between actually independant objects (like ye olde friend classes) is messy and error prone. I'd like to avoid that. I see what you mean, but I think the ad-hoc object we return here is a good thing - in a way its just an extension of the shared-cache, not really indpendant at all right?

as for memory management - closures follow the same rules as classes, but its harder (for humans, not for V8) to track the lifetimes of things in lexical scopes, vs. where its spelled out in a class. I really dont think you'd see any GC / perf improvement with a class here.

as for what is closed over by these functions, its the cache itself, `const id` and the functions themselves. not really much to worry about to my eye.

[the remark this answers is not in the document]
lanesawyer: Good thing I recently added #172 😁

I have no experience with these tools though so I don't have any concrete recommendations, but I think getting some performance testing set up for the repo will be an excellent long-term decision

packages/core/src/shared-priority-cache/priority-cache.ts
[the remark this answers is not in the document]
froyo-np: I wrote some tests - over 1million puts and 999000 evictions, the cost of a single put (which can and does call evict, that time is included in the count) is on average, less than 10 microseconds on my computer.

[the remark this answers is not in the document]
Jarbuckle: All very good points! I would argue that since all caches have to operate on a heuristic (no crystal balls here!), the most efficient management of the cache would probably be based on recent history -- like if there were a lot of evictions happening in the last, say, 100ms, then it isn't unreasonable to assume that there will be a lot more evictions happening in the near future, so evicting a larger number of entries at once to reduce the (admittedly slight) overhead would potentially be faster overall.

That said, you make a good point about how long it takes (from the system's perspective) for the fetches to complete before the eviction is needed. Better to have the data available for re-renderings for as long as possible. Arguably, evicting relatively rarely-used or low-priority items shouldn't be a major issue (assuming the users have prioritized their chunks correctly), since that's a part of what that prioritization represents, at least by my read of your design.

With all that said, the numbers speak for themselves 🙂 Seeing as evictions appear to be very minimal in their performance impact, there's very little to worry about!

packages/core/src/shared-priority-cache/shared-cache.ts
froyo-np: I've re-written this - now I'm going to try to write some tests for it

packages/core/src/shared-priority-cache/shared-cache.ts
froyo-np: renamed

site/src/examples/common/react/priority-cache-provider.tsx
```diff
@@ -0,0 +1,54 @@
+import { logger, SharedPriorityCache } from '@alleninstitute/vis-core';
+import { createContext, useEffect, useRef, type PropsWithChildren } from 'react';
+import REGL from 'regl';
+
+export const SharedCacheContext = createContext<{
+    regl: REGL.Regl;
+    cache: SharedPriorityCache;
+} | null>(null);
+
+export function SharedCacheProvider(props: PropsWithChildren) {
```
lanesawyer, line 10: Found the issue! On first render, `state` is `undefined` and then it's not rendering again because nothing is happening. Hence why it pops up on reload.

Instead, we can do it like this, and it all renders!

```
export function SharedCacheProvider(props: PropsWithChildren) {
    const [state] = useState(() => {
        const newState = new StateHelper(2000 * 1024 * 1024, 50, ['oes_texture_float']);
        logger.info('server started...');
        return newState;
    });
    
    const { children } = props;
    
    useEffect(() => {
        return () => {
            logger.info('shared cache disposed');
            state.destroy();
        };
    }, [state]);
    
    return (
        <SharedCacheContext.Provider
            value={{ regl: state.regl, cache: state.cache }}
        >
            {children}
        </SharedCacheContext.Provider>
    );
}
```

I totally forgot the `useState` initializing funciton was a thing... We can clean up SO MANY `useRef`s with that now!

site/src/examples/omezarr/render-utils.ts
```diff
@@ -0,0 +1,124 @@
...
+    const renderer = buildOmeZarrSliceRenderer(regl, decoder);
+    const client = cache.registerClient<Thing, Record<string, Tex>>({
+        cacheKeys: (item) => {
+            const channelKeys = Object.keys(item.settings.channels);
+            return channelKeys.reduce((chans, key) => {
+                return { ...chans, [key]: renderer.cacheKey(item.tile, key, item.dataset, item.settings) };
+            }, {});
+        },
+        fetch: (item) => {
+            // the Renderer<...> type obscures the fact that these are always cached textures... TODO  fix that?
```
lanesawyer, line 56: A TODO to fix?

site/src/examples/omezarr/render-utils.ts
```diff
@@ -0,0 +1,124 @@
...
+            ),
+        onDataArrived: onData,
+    });
+    // let dbl: BufferPair<REGL.Framebuffer2D> = {
+    //     writeTo: regl.framebuffer({ width: screenSize[0], height: screenSize[1] }),
+    //     readFrom: regl.framebuffer({ width: screenSize[0], height: screenSize[1] }),
+    // };
+    const [width, height] = screenSize;
+    const target: REGL.Framebuffer2D = regl.framebuffer({ width, height });
+    return {
```
lanesawyer, line 78: Woooo loving this. More straightforward for sure!

lanesawyer (commented):
Alright, a variety of comments. Some are minor items to fix, some are questions, some are me being happy to see cool stuff, and one is how to get them all rendering on the first page load!

site/src/examples/common/react/priority-cache-provider.tsx
lanesawyer: After sleeping on it, I'm not sure that `useState` is the best option, since the render server is one of those things we don't want being deleted or recreated with any of React's lifecycles.

To keep it a `useRef` but fix it, we just take the `state.current =` stuff out of the use effect and do an `if` statement checing if it's null and setting it if it is.

packages/core/src/shared-priority-cache/priority-cache.ts
```diff
@@ -0,0 +1,162 @@
...
+    }
+    private beginFetch({ key, fetch }: PendingResource) {
+        const abort = new AbortController();
+        this.pendingFetches.set(key, abort);
+        return fetch(abort.signal)
+            .then((resource) => {
+                this.put(key, resource);
+            })
+            .catch((_reason) => {
+                //ignore
```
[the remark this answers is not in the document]
froyo-np: this is a good callout - because the promise doesnt get created until a thing moves through a queue, the fetch is faar removed from the thing that wants the result, so I couldnt find a good way to get this error back to the part of the code that should be responsible for it...

packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,134 @@
...
+
+export type ClientSpec<Item, ItemContent extends Record<string, Resource>> = {
+    isValue: (v: Record<string, Resource | undefined>) => v is ItemContent;
+    cacheKeys: (item: Item) => { [k in keyof ItemContent]: string };
+    onDataArrived?: (cacheKey: string) => void; // todo very not helpful
+    fetch: (item: Item) => { [k in keyof ItemContent]: (abort: AbortSignal) => Promise<Resource> };
+};
+type ObjectValue<T extends Record<string, any>> = T extends Record<string, infer Value> ? Value : never;
+
+type KV<T extends Record<string, any>> = readonly [keyof T, ObjectValue<T>];
```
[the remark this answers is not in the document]
froyo-np: nice!

packages/core/src/shared-priority-cache/priority-cache.ts
```diff
@@ -0,0 +1,162 @@
...
+            })
+            .finally(() => {
+                this.pendingFetches.delete(key);
+                this.fetchToLimit();
+            })
+            .then(() => this.notify?.(key));
```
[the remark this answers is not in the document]
froyo-np: but, I've re-written this anyway and taken your suggestion along the way. should be fine

packages/core/src/shared-priority-cache/shared-cache.ts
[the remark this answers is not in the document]
froyo-np: unknowns work great - not sure what was up when I switched away from them

packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,134 @@
...
+                const keys = spec.cacheKeys(item);
+                Object.entries(spec.fetch(item)).forEach(([sk, fetcher]) => {
+                    const ck = keys[sk];
+                    if (ck !== undefined) {
+                        this.cache.enqueue(ck, fetcher);
+                    }
+                });
+            }
+        };
+        const setPriorities = (low: Iterable<Item>, high: Iterable<Item>) => {
```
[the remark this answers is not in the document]
lanesawyer: Good points. I'm happy with just the two for now since we don't have a reason for more at the moment. Can be a future-us problem if we want!

site/src/examples/common/react/priority-cache-provider.tsx
froyo-np: hmmm - this does work, but I'm noticing it seems to rebuild itself every frame... surely I did something wrong

froyo-np: somehow the zoomHandler was constantly a new object - anyway fixed it this is working well now

packages/core/src/shared-priority-cache/priority-cache.test.ts
```diff
@@ -0,0 +1,213 @@
...
+            expect(putOverheadMS / onemil).toBeLessThan(0.01); // yup, we are expecting this call to take less than 10 microseconds on average.
+            // a photon travels about 10,000 feet in that time
+            expect(numEvicted).toBe(999000);
+            expect(rePrioritizeEvents).toBe(onemil / 100);
+            expect(rePrioritizeOverheadMS / rePrioritizeEvents).toBeLessThan(0.1);
+        },
+        { timeout: 10000 },
+    );
+    test(
+        'enqueue with instant fetching - overall speed',
```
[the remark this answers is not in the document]
froyo-np: regarding the timings and expectations of these - they are completely dependant on the machine on which you run them, and the circumstances of that machine! they're easy to make, and they can help track down or think through bottlenecks... but they are pretty hard to use as something like a perf/regression test

packages/core/src/shared-priority-cache/priority-cache.test.ts
```diff
@@ -0,0 +1,213 @@
...
+            cache.put(ID, item);
+            putOverheadMS += performance.now() - begin;
+        }
+        console.log('1 million puts,', numEvicted, 'evictions,', putOverheadMS, 'ms total');
+        console.log('each put (avg ms): ', putOverheadMS / onemil);
+        expect(putOverheadMS / onemil).toBeLessThan(0.001); // yup, we are expecting this call to take less than one microsecond on average.
+        // a photon travels about 1000 feet in that time
+        expect(numEvicted).toBe(999000);
+    });
+    test(
```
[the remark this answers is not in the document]
froyo-np: fixed

packages/core/src/shared-priority-cache/priority-cache.ts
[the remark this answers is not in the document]
froyo-np: resolved

packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -0,0 +1,134 @@
+import * as lo from 'lodash'; // vite/astro went bonkers - not sure why this helps, core has always had lodash in it...
```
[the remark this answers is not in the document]
froyo-np: I was having very weird build system issues, but I think the actual trouble was resolved by your "type":"module" fix in package.json a few days ago. This has now been cleaned up!

lanesawyer (approved):
Looking good! One question that I think was an oversight on a refactor?

I'm also seeing lots of the "AbortError: the operation was aborted" and "uncaught exception: cancelled" in the dev console. It would be nice to catch those and log them with our logger if they're needed, or just suppress them if they're part of the normal flow (which is my assumption, since we're aborting things a lot more smartly now)

The error clean up doesn't _have_ to happen now since it's not integrated with any apps yet, but it'd be nice to not fill up the console with errors if we don't need to

packages/core/src/shared-priority-cache/priority-cache.test.ts
```diff
@@ -189,3 +75,135 @@ describe('basics', () => {
...
+        const fakeStore: FakeStore = new FakeStore();
+        const fakeFetchItem = (id: string) => (_sig: AbortSignal) => {
+            return promises.promiseMe(() => factory.create(id, 33));
+        };
+        factory = new PayloadFactory();
+        const priorities: Record<string, number> = {};
+        const score = (k: string) => priorities[k] ?? 0;
+        const cache: PriorityCache = new PriorityCache(fakeStore, score, 1000, 20);
+
+        const onemil = 100_000;
```
Jarbuckle, line 183: Maybe rename this const, since it's not a million here?

packages/core/src/shared-priority-cache/priority-cache.ts
```diff
@@ -84,15 +85,15 @@ export class PriorityCache {
...
+                this.notify?.(key, { status: 'success' });
             })
-            .catch((_reason) => {
-                //ignore
+            .catch((reason) => {
+                this.notify?.(key, { status: 'failure', reason });
             })
             .finally(() => {
                 this.pendingFetches.delete(key);
                 this.fetchToLimit();
```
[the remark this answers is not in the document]
froyo-np: we do - on line 88, which you cant see in this diff :)

PR #180, opened 2025-07-17
chore: Update release numbers for release

lanesawyer:
See release notes for details

`core` was already bumped in the most recent `core` PR

And we don't need to have lockfile changes because `pnpm` doesn't put the version number in them!


Jarbuckle (approved):
Numbers go brrrrrr :)

PR #187, opened 2025-08-14
Allow slices from planes by % rather than slice index - as the number of slices is not constant at all scales in most volumes

froyo-np:
# 

# What
The renderer accepts settings that included "orthoVal" which is an index of a plane in a volume (along the axis that is orthagonal to the plane itself). that is fine, but troublesome for volumes which have different sizes in that axis at different scales.
this PR adds the option to pick the same value via a percent value, which will cause the image to be stable while zooming, and also avoid accidental index-out-of-bounds errors
# How

# Screenshots

# PR Checklist

### DIFF
--- a/site/src/examples/omezarr/sliceview.tsx
+++ b/site/src/examples/omezarr/sliceview.tsx
@@ -36,7 +36,7 @@ const settings: RenderSettings = {
         },
     },
     plane: PLANE_XY,
-    planeLocation: { index: 3 },
+    planeLocation: 0.5,
     camera: {
         // the omezarr renderer expects a box in whatever space is given by the omezarr file itself in its
         // axes metadata = for example, millimeters. if you load a volume that says its 30mm X 30mm X 10mm,


packages/omezarr/src/sliceview/loader.ts
```diff
@@ -94,15 +94,26 @@ export function getVisibleTiles(
         screenSize: vec2;
     },
     plane: CartesianPlane,
-    orthoVal: number,
+    planeLocation:
```
froyo-np, line 97: the meat of the change - orthoVal was always assumed to be the voxel-index of a slice along the given plane. I renamed it to be more descriptive, and it is now a union to support either index or parameter (percentage) specification

site/src/examples/omezarr/omezarr-demo.tsx
```diff
@@ -92,7 +101,7 @@ export function OmezarrDemo() {
         loadMetadata(res).then((v) => {
             setOmezarr(v);
             setOmezarrJson(JSON.stringify(v, undefined, 4));
-            setPlaneIndex(Math.floor(v.maxOrthogonal(PLANE_XY) / 2));
+            setPlaneIndex(0.5);
```
froyo-np, line 104: ok I seem to have committed some stuff I meant to revert - I'll put this back

site/src/examples/omezarr/omezarr-demo.tsx
```diff
@@ -43,6 +43,15 @@ const demoOptions: DemoOption[] = [
             url: 's3://allen-genetic-tools/tissuecyte/823818122/ome_zarr_conversion/823818122.zarr/',
         },
     },
+    {
+        value: 'opt5',
```
lanesawyer, line 47: We need to either switch "V3 Zarr Example Image (S3) (color channels: [R, G, B])" to "opt6" or move this one to the end of the list and name it "opt6". Right now the OME-Zarr V3 doesn't work because the key values are clashing

packages/omezarr/src/zarr/loading.ts
```diff
@@ -179,10 +179,27 @@ export function pickBestScale(
     }, datasets[0]);
     return choice ?? datasets[datasets.length - 1];
 }
-
+// TODO this is a duplicate of indexOfDimension... delete one of them!
```
lanesawyer, line 182: Let's take care of this unification now. `indexOfDimension` isn't exported in the `index` file (and we don't use it outside the package anyway) so pick which name you like best and we'll go with it!

lanesawyer (commented):
Type change and subsequent logic make sense. Got a couple small items, only the `value` dropdown index change one is _truly_ necessary at this point so we don't have a broken demo page.

site/src/examples/omezarr/omezarr-demo.tsx
froyo-np: sorry about that - fixed!

packages/omezarr/src/zarr/loading.ts
froyo-np: done

PR #200, opened 2025-10-03
chore: Adding in a caching multithreaded Fetch Store

Jarbuckle:
# chore: Adding in a caching multithreaded Fetch Store
## Performance Enhancements for V3 OME-Zarr Viewing: Part II

# What

In order to enhance loading performance for V3 OME-Zarr files in the image viewer, a new Fetch Store is needed that can both a) delegate the requests for specific data chunks to multiple child threads/workers, and b) cache the results so that excess requests are not sent at all.

**NOTE:** This PR does not include any actually-testable functionality, it is purely a code PR at this stage. The testing will come in the next PR where these classes are put to use loading OME-Zarr images into a renderer.

# How

- Added a `CachingMultithreadedFetchStore` class that extends the Zarrita `FetchStore`
- Added a `WorkerPool` class that generalizes earlier implementations by Noah
- Added a `fetch-slice.worker` script and `fetch-slice.interface` module that define a Worker that specifically knows how to fetch Zarr slices using a private `FetchStore`; this `FetchStore` is initialized with the same configuration used to generate the `CachingMultithreadedFetchStore` that owns these workers (via its `WorkerPool` instance)

# PR Checklist


packages/omezarr/src/zarr/cached-loading/fetch-slice.interface.ts
```diff
@@ -0,0 +1,112 @@
...
+export const CANCEL_MESSAGE_TYPE = 'cancel' as const;
+
+export type FetchSliceMessage = {
+    type: typeof FETCH_SLICE_MESSAGE_TYPE;
+    id: string;
+    payload: FetchSliceMessagePayload;
+};
+
+export type FetchSliceResponseMessage = {
+    type: typeof FETCH_SLICE_RESPOSNE_MESSAGE_TYPE;
```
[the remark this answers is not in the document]
Jarbuckle: It's just as a way of avoiding having to define the value of it in more than one place, in case the value changes in the future.

site/src/examples/omezarr/render-utils.ts
```diff
@@ -48,8 +50,9 @@ export function buildConnectedRenderer(
     const client = cache.registerClient<Thing, Record<string, Tex>>({
         cacheKeys: (item) => {
             const channelKeys = Object.keys(item.settings.channels);
-            return channelKeys.reduce((chans, key) => {
-                return { ...chans, [key]: renderer.cacheKey(item.tile, key, item.dataset, item.settings) };
```
froyo-np, line 52: are you making this changes for performance reasons?

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,187 @@
...
+        this.#priorityMap.set(cacheKey, Date.now());
+        return new Uint8Array(cached.buffer());
+    }
+
+    async #doFetch(
+        key: zarr.AbsolutePath,
+        range: zarr.RangeQuery | undefined,
+        options: TransferableRequestInit,
+    ): Promise<Uint8Array | undefined> {
+        const response = await this.#workerPool.submitRequest(
```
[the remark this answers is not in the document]
froyo-np: because you have a pool that returns a promise, I think you could use the AsyncPriorityCache here right? you just have to wrap the promise here in a fetcher function (no args, if I remember right) and then it will handle annoying stuff like repeat requests (that are not yet fulfilled) for you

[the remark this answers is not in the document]
Jarbuckle: Ah, yes, I need to make sure that any `AbortSignal` from the `get`/`getRange` calls is handled properly here, good call!

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,187 @@
...
+        return this.#priorityMap.get(key) ?? 0;
+    }
+
+    #fromCache(cacheKey: CacheKey): Uint8Array | undefined {
+        const cached = this.#dataCache.get(cacheKey);
+        if (cached === undefined) {
+            return undefined;
+        }
+        this.#priorityMap.set(cacheKey, Date.now());
+        return new Uint8Array(cached.buffer());
```
froyo-np, line 134: this can throw an exception... I think its almost certainly impossible, but with async stuff it can be hard to tell. The purpose of this cache is a little different from what I had in mind when I set it up for WebGL resource management... I'm wondering if its safer to simply have this Cacheable's destroy() function do nothing?

site/src/examples/omezarr/render-utils.ts
froyo-np: I personally think the linter is full of **it, but I suppose these changes are fine

packages/omezarr/src/zarr/cached-loading/store.ts
Jarbuckle: Hmm, okay; is it the call to `cached.buffer()` that can throw an exception?

Jarbuckle: Done!

packages/omezarr/src/zarr/cached-loading/store.ts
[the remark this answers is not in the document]
Jarbuckle: Done!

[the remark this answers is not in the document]
Jarbuckle: Done (per below)!

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,256 @@
...
+     * Maintains a pool of available worker threads.
+     */
+    #workerPool: WorkerPool;
+
+    /**
+     * Stores the current set of cached data that has been successfully
+     * fetched. This data is stored in raw byte array form so that it
+     * integrates properly with the Zarrita framework.
+     */
+    #dataCache: AsyncPriorityCache<CacheableByteArray>;
```
froyo-np, line 115: well, I know I asked for this change (asyncPriorityCache instead of priorityCache) but now I think I dont really like it 😅

but! I think I have learned why this feels a bit off:

1. AsyncPriorityCache does indeed handle multiple requests, etc but its specifically designed to accept a list of requests, and then never respond (via promises) to those requests - the expectation is that you just ask for them again later and magically, there they are. 
2. AsyncPriorityCache uses a queue to limit fetches - thats nice, but not nice for our store, right? we want a simple interface `gimmieData(x)==>Promise<Stuff>`

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * Maps cache keys to numeric times; the higher the time, the higher the priority.
+     *
+     * This effectively means that more frequently-requested items will be kept longer.
+     */
+    #priorityMap: Map<CacheKey, number>;
+
+    #pendingRequests: Map<CacheKey, PendingRequest<Uint8Array | undefined>>;
```
froyo-np, line 124: this is the part felt fishy to me - AsyncPriorityCache has its own map of pending requests! that made me realize, we're doing most of its work here already, so much so that its not really helping

packages/omezarr/src/zarr/cached-loading/store.ts
froyo-np: ok hear me out - if you put dataCache back to being just a PriorityCache (sorry!) - and you keep your map of pending requests, you're basically done - you can check for pending requests in get() and return the cached promise. if Not pending, doFetch() and return that promise. you can delete the dataRecieved fn, and I think all the score stuff just works and is fine already.

packages/core/src/shared-priority-cache/priority-cache.ts
```diff
@@ -14,43 +15,54 @@ export interface Store<K extends {}, V> {
...
 
-function negate(fn: (k: CacheKey) => number) {
+function negate(fn: ScoreFn) {
     return (k: CacheKey) => -fn(k);
 }
+
 export type FetchResult = { status: 'success' } | { status: 'failure'; reason: unknown };
 
 export class PriorityCache<T extends Cacheable> {
-    private store: Store<CacheKey, T>;
```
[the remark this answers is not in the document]
Jarbuckle: I hear ya; it's taking some getting used to for me, too.

packages/omezarr/src/zarr/cached-loading/store.ts
Jarbuckle: Now that you lay it out, you're right, that does make sense 😅 Thanks for being a second set of eyes on this!

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,256 @@
...
+            }
+            const arr = new Uint8Array(response.payload);
+            return new CacheableByteArray(arr);
+        };
+
+        const cacheKey = asCacheKey(key, range);
+        this.#priorityMap.set(cacheKey, Date.now());
+
+        if (abort) {
+            abort.onabort = () => {
```
froyo-np, line 212: hmm - this is a pickle. Its a bit of a pickle in this version, and if you put dataCache back to just being PriorityCache (not async).

pretend unit test:
request chunk A twice, then abort one of them.

if you abort the second one - it will set the priority to 0, and then reprioritize will actually send the cancel method to the worker. I think that will have the effect of causing both promises to... never resolve, but also never be rejected right?

Jarbuckle: It's definitely a pickle. 🤔 I suppose the store could track all of the individual instances of a request for a particular cache key, and "abort"/delete them specifically if the abort signal is received for that one in particular; then, if none remain, it can do the actual cancellation of the request.

Part of the problem is what the expected interface *should* be for this. Is it reasonable to assume that the same byte range will be requested more than once for valid reasons, rather than by mistake and/or redundancy? And if one of them is aborted, should the assumption be that that range isn't needed at all, or that the specific caller doesn't need it anymore?

If it's the latter, I think something like I described above might be necessary. If a particular caller wants to actively cancel a request for data, but some other caller still wants that data, we definitely do not want to cancel it unless no remaining requestors are active.

packages/omezarr/src/zarr/cached-loading/store.ts
Jarbuckle: I've updated it to use `PriorityCache` again 🙂 Doesn't try to resolve the Abort issues we're discussing below, though; will tackle that on Monday.

packages/omezarr/src/zarr/cached-loading/store.ts
froyo-np: as far as I can tell - this cache behaves correctly. It would be great to have some unit tests to prove it out

Jarbuckle: > so I think the answer here would be to match the behavior of the regular zarrita fetch store. If you take a look at its source (its docs are a bit too wimpy) you'll see that the abortController you pass to get() gets handed off to Window.fetch(). I think that implies the following:

Hmm, in the scenario where each get() actually sends its own fetch and can therefore be stopped independently, that makes sense. In our case, we're essentially treating all requests for the same resource as the same request, so I'm not sure this would work for us. Am I missing something? 🤔

packages/omezarr/src/zarr/cached-loading/fetch-slice.interface.ts
```diff
@@ -0,0 +1,112 @@
...
+
+export type FetchSliceMessagePayload = {
+    rootUrl: string;
+    path: AbsolutePath;
+    range: RangeQuery;
+    options?: TransferrableRequestInit | undefined;
+};
+
+export type WorkerMessage = {
+    type: string;
```
froyo-np, line 17: is not the `type` here a union of the literals `BLAH_BLAH_TYPE` that you define on line 40-42?

packages/core/src/shared-priority-cache/priority-cache.ts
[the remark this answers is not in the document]
lanesawyer: Agreed. Looks gross but it is the new native way to do it lol 🤷🏻

packages/omezarr/src/zarr/cached-loading/store.ts
froyo-np: (nitpick - not important just confused :) )

Jarbuckle: Oh hang on, I think I might've made a goober here; got some ownership-transfer stuff mixed up in my head 😅

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,288 @@
...
+    #arr: Uint8Array;
+
+    constructor(arr: Uint8Array) {
+        this.#arr = arr;
+    }
+
+    destroy() {}
+
+    sizeInBytes(): number {
+        return this.#arr?.byteLength ?? 0;
```
froyo-np, line 39: hmm - not clear to me how arr could be undefined here - ah perhaps this is leftovers from a refactor here?

packages/omezarr/src/zarr/cached-loading/store.ts
```diff
@@ -0,0 +1,288 @@
...
+            },
+            isFetchSliceResponseMessage,
+            [],
+        );
+
+        request
+            .then((response: FetchSliceResponseMessage) => {
+                const payload = response.payload;
+                if (payload === undefined) {
+                    resolve(undefined);
```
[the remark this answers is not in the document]
Jarbuckle: I agree. 🤷🏽 For now, I'd like to stick with the FetchStore contract, but in the future it may make sense to break free a bit.

packages/omezarr/src/zarr/cached-loading/worker-pool.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export class WorkerPool {
+    #workers: Worker[];
+    #promises: Map<string, MessagePromise<WorkerMessageWithId>>;
+    #which: number;
+
+    constructor(size: number, workerModule: URL) {
+        this.#workers = new Array(size);
+        for (let i = 0; i < size; i++) {
+            this.#workers[i] = new Worker(workerModule, { type: 'module' });
```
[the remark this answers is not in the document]
Jarbuckle: I haven't, and it's one of the things I'm a bit concerned about, too 😓 Testing will happen soon!

packages/omezarr/src/zarr/cached-loading/fetch-slice.interface.ts
Jarbuckle: Okay, that move is now done; sent it over to `core/workers`.

packages/omezarr/src/zarr/cached-loading/store.ts
Jarbuckle: This should be fixed now, too!

packages/core/src/workers/worker-pool.ts
```diff
@@ -19,24 +19,51 @@ type MessagePromise<T extends WorkerMessageWithId> = {
...
+export enum WorkerStatus {
+    Available = 'Available',
+    Working = 'Working',
+    Unresponsive = 'Unresponsive',
+}
+
 export class WorkerPool {
     #workers: Worker[];
     #promises: Map<number, MessagePromise<WorkerMessageWithId>>;
+    #statuses: Map<number, WorkerStatus>;
```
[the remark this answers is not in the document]
Jarbuckle: I agree that doing any more on that front would indeed be a major side-quest, which is why I'm not doing any more on it right now 😉 As you say, we've got working examples of specific fetch-pools, so adding another one didn't feel like the right path forward for this.

packages/core/src/workers/worker-pool.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+export class WorkerPool {
+    #workers: Worker[];
+    #promises: Map<number, MessagePromise<WorkerMessageWithId>>;
+    #statuses: Map<number, WorkerStatus>;
+    #timeOfPreviousHeartbeat: Map<number, number>;
+    #which: number;
+
+    constructor(size: number, workerModule: URL) {
```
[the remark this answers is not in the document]
Jarbuckle: Yep, that's a very fair point! Would be good to define a way to tie these together much more securely. I'd say that's probably a feature for the future, though. Any concerns with waiting on that?

packages/core/src/workers/messages.ts
```diff
@@ -0,0 +1,25 @@
+import { z } from 'zod';
+
```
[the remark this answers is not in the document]
Jarbuckle: Zod's been removed 👍🏽

packages/core/src/workers/worker-pool.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export class WorkerPool {
+    #workers: Worker[];
+    #promises: Map<string, MessagePromise<WorkerMessageWithId>>;
+    #which: number;
+
+    constructor(size: number, workerModule: URL) {
+        this.#workers = new Array(size);
+        for (let i = 0; i < size; i++) {
+            this.#workers[i] = new Worker(workerModule, { type: 'module' });
```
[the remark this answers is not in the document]
Jarbuckle: Added a new example app that sets up a basic WorkerPool and tests that Workers are created and sending messages; all good!

PR #253, opened 2026-05-05
Slight changes to the CacheClient interface

froyo-np:
# What
 - [x] the shared priority cache is accessed via "clients" - these clients provide a way to check at runtime if content returned by the cache will satisfy the needs (enough data for the renderer to render something, for example). This function is now passed the request object, this makes this check much more polymorphic, allowing more re-use of cache-clients if the shape of data to be used changes a bit.
 - [x] update the scatterbrain library's use of the cache client to not require regl directly
 - [x] update the scatterbrain library's use of the cache client to use the new isValue check, to make it easy to change shader settings without requiring a new shader and a new cacheClient.
# How

# Screenshots

# PR Checklist


packages/core/src/shared-priority-cache/shared-cache.ts
```diff
@@ -19,7 +19,7 @@ type CacheInterface<Item, ItemContent extends Record<string, Cacheable>> = {
 };
 
 export type ClientSpec<Item, ItemContent extends Record<string, Cacheable>> = {
-    isValue: (v: Record<string, Cacheable | undefined>) => v is ItemContent;
+    isValue: (v: Record<string, Cacheable | undefined>, item: Item) => v is ItemContent;
```
froyo-np, line 22: this is the api change - its non-breaking. All prior cache-clients that pass an isValue method that ignores the new argument are valid, will work at runtime, and will pass typechecking still.

packages/scatterbrain/src/cache-client.ts
```diff
@@ -0,0 +1,65 @@
+import type { Cacheable, SharedPriorityCache } from '@alleninstitute/vis-core';
+import reduce from 'lodash/reduce';
+import type { WebGLSafeBasicType } from './typed-array';
+import type { ColumnRequest, Item, } from './types';
+
+
+type Content<V extends Cacheable> = Record<string, V>
+
+export function buildScatterbrainCacheClient<V extends Cacheable>(
```
froyo-np, line 8: nearly identical to the version that has been deleted from ./renderer.ts - this version is simply slightly more generic.
it does not need to know about regl, nor does it rely on a closure over the set of requested columns to validate its contents prior to serving up data.

packages/scatterbrain/src/renderer.ts
```diff
@@ -1,85 +1,21 @@
...
+
 export type Item = Readonly<{
     dataset: SlideviewScatterbrainDataset | ScatterbrainDataset;
     node: TreeNode;
     bounds: box2D;
     columns: Record<string, ColumnRequest>;
 }>;
-type Content = Record<string, VBO>;
-
-export function buildScatterbrainCacheClient(
```
froyo-np, line 18: see cache-client.ts

TheMooseman (approved):
Works on my machine, can't find anything to complain about

PR #256, opened 2026-05-11
feature: WebGPU Rendering Enhancements

Jarbuckle:
# feature: WebGPU Rendering Enhancements

## What

In support of the Connectomics initiative, we are making updates to Vis to enable WebGPU rendering -- initially for the Scatterbrain package, but not limited to that package. WebGPU rendering will allow us to render to multiple canvases, as well as situating Vis for long-term support of modern web technologies.

This PR is focused on building reusable systems to simplify the process of defining and using WebGPU rendering pipelines. It will be an ongoing draft PR until we have settled on the systems that we want (R&D process).

## How

- Built programmatic wrappers around shader elements to make building shaders a modular and reusable system.
- MORE TBD

## PR Checklist


packages/scatterbrain/src/render/webgpu/shaders/shader.ts
```diff
@@ -0,0 +1,34 @@
...
+ * This file defines the `WGSLShader` class, which represents a shader program in WGSL. 
+ * It includes methods for serializing and deserializing the shader definition, as well 
+ * as generating the WGSL source code from the defined declarations. The `shader` 
+ * function is a simple helper function for creating a new `WGSLShader` instance from an
+ * array of declarations.
+ */
+
+import type { Declaration } from "./declarations";
+
+export class WGSLShader {
```
froyo-np, line 11: still reading through all this - so far I like a lot of it!

my first thought here though is that I dont think this being a class is helpful!
why not `type WGSLShader = Declaration[]` ?

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
```diff
@@ -0,0 +1,34 @@
...
+        this.#definition = { declarations };
+    }
+
+    static deserialize(serialized: string): WGSLShader {
+        const definition = JSON.parse(serialized);
+        return new WGSLShader(definition.declarations);
+    }
+
+    serialize(): string {
+        return JSON.stringify(this.#definition);
```
froyo-np, line 24: if this class was instead an open type - then its users could do their own serialization in whatever form they wanted

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
```diff
@@ -0,0 +1,34 @@
...
+import type { Declaration } from "./declarations";
+
+export class WGSLShader {
+    #definition: { declarations: Declaration[] };
+
+    constructor(declarations: Declaration[]) {
+        this.#definition = { declarations };
+    }
+
+    static deserialize(serialized: string): WGSLShader {
```
froyo-np, line 18: this is just asking for runtime errors imho

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
```diff
@@ -0,0 +1,34 @@
...
+        const definition = JSON.parse(serialized);
+        return new WGSLShader(definition.declarations);
+    }
+
+    serialize(): string {
+        return JSON.stringify(this.#definition);
+    }
+
+    asSource(): string {
+        return this.#definition.declarations.map((d) => d.__gen()).join(';\n');
```
froyo-np, line 28: this is the only useful part of this class - could easily be
`function asWgslSource(declaration[]) { ... }`

packages/scatterbrain/src/render/webgpu/shader.ts
```diff
@@ -0,0 +1,298 @@
...
+    view: vec4;
+    spatialFilterBox: vec4;
+    filteredOutColor: vec4;
+    highlightColor: vec4;
+    screenSize: vec2;
+    offset: vec2;
+    highlightValue: number;
+} & QuantitativeFilterRanges;
+
+export const applyCamera = func(
```
froyo-np, line 58: I like a bunch of the other constructed wrappers, but I'm not so sure about the func constructor. it feels less ergonomic then just the literal wgsl string you'd use - and because its internal to a shader, its not really buying us much in terms of type safety is it? functions defined this way still get invoked in raw WGSL

froyo-np: I'm not saying get rid of it - but what if Declaration had an escape hatch member type to allow a literal blob of WGSL?

packages/scatterbrain/src/render/webgpu/shader.ts
```diff
@@ -0,0 +1,298 @@
...
+    const vertexStruct = makeVertexStruct(config);
+    const uniformStruct = makeUniformStruct(config);
+
+    return shader([
+        vertexStruct,
+        uniformStruct,
+        vsOutputStruct,
+        uniform('unis', uniformStruct.name, 0, 0),
+        texture(categoricalTable, 'texture_2d<f32>', 0, 1),
+        texture(gradientTable, 'texture_2d<f32>', 0, 2),
```
froyo-np, line 151: this part I do like - its clear to see the structure of the things in the 'header' of the shader are generated, and because the gross part of constructing them is done elsewhere, this is a nice concise reading of the interface to the shader from outside.

packages/scatterbrain/src/render/webgpu/shader.ts
```diff
@@ -0,0 +1,298 @@
...
+        applyCamera,
+        rangeParameter,
+        within,
+        constant('clip', /*wgsl*/ `array<vec2f,4>(vec2f(1, -1), vec2f(1, 1), vec2f(-1, -1), vec2f(-1, 1))`),
+        vertexEntry(
```
froyo-np, line 156: here again - I think the preamble to the vertex main fn in raw wgsl is much more legible than this:
`@vertex fn vmain(v:Vertex)->VsOutput`

packages/scatterbrain/src/render/webgpu/shaders/declarations.ts
```diff
@@ -0,0 +1,405 @@
...
+        __identType: 'struct',
+        name,
+        fields,
+        __gen: () => `struct ${name} { ${fields.map((f) => f.__gen()).join(', ')} }`,
+    };
+}
+
+export function param(
+    name: string,
+    type: string,
```
froyo-np, line 321: might be nice to constrain this more - perhaps (if its a structDeclaration) we could pull the .name out so the person authoring these doesnt have to

packages/scatterbrain/src/render/webgpu/shaders/declarations.ts
```diff
@@ -0,0 +1,405 @@
...
+        group,
+        binding,
+        ...(accessMode !== undefined && { accessMode }),
+        ...(attributes !== undefined && { attributes }),
+        __gen: () =>
+            `${renderAttrs(attributes)}@group(${group}) @binding(${binding}) var<storage${accessMode !== undefined ? `, ${accessMode}` : ''}> ${name}: ${type}`,
+    };
+}
+
+export function member(name: string, type: string, attributes?: VariableOrValueAttribute[]): StructMemberDeclaration {
```
froyo-np, line 301: because the type parameter here is just a raw string, and I dont have syntax highlighting because I'm not writing raw WGSL , I feel like this is a step backward in type-safeness, right?

packages/scatterbrain/src/render/webgpu/shader.ts
Jarbuckle: Does it need to be internal to a shader, though? 🤔 The goal behind this was to provide a way of making function declarations reusable across shaders; allowing for a simple "chunk of WGSL" approach feels much less encapsulated (and I do love me some encapsulation, since it sets expectations -- crucial for the programmer-program interface, in my mind).

packages/scatterbrain/src/render/webgpu/shader.ts
Jarbuckle: With that, I do agree. Part of the lack of legibility is due to Oxfmt's default formatting rules (insisting on having parameters on separate lines -- not really helping in this case), but yes, it's more readable in raw WGSL.

packages/scatterbrain/src/render/webgpu/shaders/declarations.ts
Jarbuckle: Yeah, having the `type` be able to specifically reference a previously-defined `struct` would be a nice addition, agreed.

packages/scatterbrain/src/render/webgpu/shaders/declarations.ts
Jarbuckle: It is, for sure. Per my response above, making this able to handle pre-defined types seems like a definite improvement.

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
Jarbuckle: It all comes down to preference, I guess 😅 I like to think of shaders (in their final form) as a definite _thing_, not just raw data.

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
Jarbuckle: True, but they could also extend the class and override the functions that way. Again, I think it all comes down to expectations and mental models; there is no functional difference, in my mind, between the model you're proposing and the one in the code currently.

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
Jarbuckle: Curious which runtime errors you're thinking are most likely? Deserialization failures? Or issues with the shader not being a valid shader? Or something else?

packages/scatterbrain/src/render/webgpu/shaders/shader.ts
Jarbuckle: True; personally, I prefer the `.asSource()` call, since to me at least it presents a clearer entity model for my mental map of the interface.

Jarbuckle:
Closed pull request in favor of the more narrowly focused #259

PR #259, opened 2026-05-14
feature: WebGPU shader builder for Vis Core [DC-9506]

Jarbuckle:
# feature: WebGPU shader builder for Vis Core [[DC-9506](https://alleninstitute.atlassian.net/browse/DT-9506)]

# What

Follow-on from an earlier (now closed) PR: #256 

That PR was a combination of different work, so the goal with this PR is to have it focused just on the WebGPU rendering pipeline, specifically the first implemented piece of that pipeline: the programmatic creation of shaders.

# How

- Created a new `webgpu` folder within `core`
- Moved over the baseline shader builder functionality from the preceding PR
- Closed the preceding PR (its other changes were inherited from work Noah's doing in another PR)
- Made some adjustments to respond to feedback provided on the preceding PR (specifically: removed the usage of classes when defining shaders, and made `StructDeclaration`s into valid values for most `type` fields -- more changes forthcoming!)

# PR Checklist


chrisj:
Could you show an example of using the builder?

Jarbuckle:
> Could you show an example of using the builder?

@chrisj This now exists! It's in the `example.test.ts` file.

Just to be clear, the actual shader logic used in that example is based on an earlier version of the point renderer that Noah wrote for his PR. Not sure if it actually _works_ at this point (Claude got its mitts on it for a bit there), but it should provide a good demonstration of the shader builder in action!

packages/core/src/webgpu/shaders/shader.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * This file defines the `WgslShader` type, which represents a shader program in WGSL.
+ * Also included are methods for serializing and deserializing the shader definition, as
```
froyo-np: this comment is a mild fib, because I was a grump about the serialize / deserialize methods mentioned here :)

packages/core/src/webgpu/shaders/example.test.ts
```diff
@@ -0,0 +1,226 @@
...
+    const within = func(
+        'within',
+        [param('v', 'f32'), param('range', 'vec2f')],
+        () => /*wgsl*/ `
+            return step(range.x,v)*step(v,range.y);
+        `,
+        returns('f32')
+    );
+
+    const sh = shader([
```
froyo-np, line 118: this example is an excellent way to see what this library is about, thanks for doing this!

packages/core/src/webgpu/shaders/example.test.ts
```diff
@@ -0,0 +1,226 @@
...
+        [param('v', 'f32'), param('range', 'vec2f')],
+        () => /*wgsl*/ `
+            return (v-range.x)/(range.y-range.x);
+        `,
+        returns('f32')
+    );
+    const within = func(
+        'within',
+        [param('v', 'f32'), param('range', 'vec2f')],
+        () => /*wgsl*/ `
```
froyo-np, line 112: can you remind me why this is a function that returns a string, and not just the string itself? I feel like we talked about it but perhaps the answer has leaked out of my skull

packages/core/src/webgpu/shaders/declarations.ts
```diff
@@ -0,0 +1,441 @@
...
+
+export type StructMemberDeclaration = IdentifierDeclaration &
+    DeclarationGenerator & {
+        type: TypeIdentifier;
+        attributes?: VariableOrValueAttribute[];
+    };
+
+export type StructDeclaration = IdentifierDeclaration &
+    DeclarationGenerator & {
+        __identType: 'struct';
```
froyo-np, line 45: so there's two of these in here, `__gen()` and `__identType`. 
In the case of __gen, I'm assuming you're trying to discourage the use of this function.

as for __identType, I dont personally see the harm in ditching the underscores - `type` is out because some of these are about WGSL types, but `kind` or `TAG` might be more pleasant to type / read?

just a personal preference nitpick (I really hate typing underscore) so feel free to ignore

packages/core/src/webgpu/shaders/declarations.ts
```diff
@@ -0,0 +1,441 @@
...
+        fields: StructMemberDeclaration[];
+    };
+
+export type AliasDeclaration = IdentifierDeclaration &
+    DeclarationGenerator & {
+        __identType: 'alias';
+        aliasedType: TypeIdentifier;
+    };
+
+export type TypeIdentifier = string | StructDeclaration | AliasDeclaration;
```
froyo-np: while we are here - we have a string as a valid TypeIdentifier - thats fine, because we dont want to enumerate all the valid wgsl built-in types - yet, right?

might not be the worst to hint at that intention:
`type WgslType = string; // TODO: enumerate builtins someday, eg. 'vec2i' | 'vec3f' ...`
`export type TypeIdentifier = WgslType | StructDeclaration | AliasDeclaration;`

froyo-np (approved):
tests seem thorough, example is good for understanding what this provides, a few nitpicks but seems worth merging to me!

packages/core/src/webgpu/shaders/example.test.ts
Jarbuckle: Yeah, the idea here was mostly to try and convince the formatter that this should be indentable so that it's easier to read 😅 Though having it be a function may potentially be helpful, maybe. 🤷🏽 I'm okay with not having it be a function, but figured the benefits (mild though they are) outweighed the drawbacks.

packages/core/src/webgpu/shaders/declarations.ts
Jarbuckle: I tend to use underscores to indicate to devs "please don't use this unless you need to" -- a way of communicating a potentially unreliable interface. I'll leave them for now, but objection duly noted (and understood)!

packages/core/src/webgpu/shaders/declarations.ts
Jarbuckle: Yep, that's the reason; I'll go ahead and include said hint as a comment, though 🙂

PR #279, opened 2026-07-15
feat: Shader building now supports Vertex Inputs [DT-9715]

Jarbuckle:
# feat: Shader building now supports Vertex Inputs [[DT-9715](https://alleninstitute.atlassian.net/browse/DT-9715)]

# What

Upgrades were made to the Shader Builder as part of the overarching effort to develop a generalized WebGPU rendering engine. These updates include:

- Addition of new Vertex Input interface declarations, enabling programmatic interactions with the parameters to the vertex shader entrypoint and vertex buffer layout automation
- Improved compile-time type-safety with the introduction of defined WGSL type declarations
- Some general cleanup to make the Shader builder module more flexible as a module (such as decoupling the `Declaration` union from the definition of `WgslShader`).
- Addition of WGSL-specific type schemas supported by Zod
- Addition of a `foundation` module that provides certain basic capabilities and a set of "semi-native" WebGPU types that can be cross-compared with whichever ambient library is supplying the WebGPU types natively in the environment, enabling compile-time detection of mismatches in WebGPU typing (a real concern in these relatively-early-days of WebGPU adoption and revision).
- Added a Docs page for the WebGPU Shader Builder, with guides and examples.

# PR Checklist

- [x] Have you smoke tested the example applications?
- [x] Have you tested the application on these browsers?
    - [x] Chrome (Fully supported)

[DT-9715]: https://alleninstitute.atlassian.net/browse/DT-9715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


site/src/content/docs/webgpu/shader-builder.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+1. **Every declaration knows how to render itself.** Each builder returns an object with a
+   `__gen()` method that emits its own slice of WGSL.
+2. **A shader is just a list of declarations.** `shader([...])` stamps the list with a stable
+   `id` and hands you back a `WgslShader`.
+3. **`asSource()` turns it into text.** It calls `__gen()` on each declaration and joins the
+   results with newlines.
+
+```ts
+import { $d, $a, $t, shader, asSource, vertexInput } from '@alleninstitute/vis-core';
```
chrisj: this might be the best solution but if it hasn't been already discussed, I think single character variable names should be used sparingly.

After reading the explanation below, I think it does read well when looking at the examples and seems well thought out

Jarbuckle: Yeah, hopefully this will be a temporary solution, or at least only needed in certain scenarios. Setting up the bundling correctly so that imports are more flexible for sub-modules of a package takes some careful consideration, and for now I felt that was out of scope.

Jarbuckle: Alternatively, we could set up some slightly-longer var names (like `$t` -> `types`, `$d` -> `decls`, `$a` -> `attrs`) and let users decide if they want to alias them shorter or just use them directly.

site/src/content/docs/webgpu/shader-builder.mdx
lanesawyer, line 1: Great doc! Started here in hopes it would orient the rest of my PR review and I was not disappointed!

lanesawyer (approved):
I did not check all the WebGPU type stuff, going to mostly rely on usage to validate those types as we go, it _looks_ comprehensive and generally correct!

Things are arraigned well, barrel file exports are used to properly control the public API, it's got passing tests, good comments, and a great high-level documentation file in Starlight.

LGTM!

site/src/content/docs/webgpu/shader-builder.mdx
```diff
@@ -0,0 +1,365 @@
...
+@group(0) @binding(1) var albedoSampler: sampler;
+@fragment fn fs_main(@location(0) uv: vec2f) -> @location(0) vec4f { return textureSample(albedo, albedoSampler, uv); }
+```
+
+</details>
+
+## Level 5 — Config-driven shaders
+
+The real payoff: because declarations are plain data, you can **generate** them. Here a single
+function turns a runtime configuration into a complete shader whose vertex layout, uniform block,
```
froyo-np, line 282: this is the payoff, and its a good one! still, it might be good to lay out some of the limitations.
for example, you can check for location-slot collisions, but you cant verify that the names in a function body match any of the fields in the structures you might have passed in, nor can you check the types for correctness, right?

site/src/content/docs/webgpu/shader-builder.mdx
```diff
@@ -0,0 +1,365 @@
...
+struct VsOut { @builtin(position) position: vec4f, @location(0) color: vec4f }
+@group(0) @binding(0) var<uniform> unis: Uniforms;
+@vertex fn vs_main(v: Vertex) -> VsOut { var out: VsOut; out.position = vec4f(v.xy, 0.0, 1.0); out.color = vec4f(1.0); return out; }
+@fragment fn fs_main(v: VsOut) -> @location(0) vec4f { return v.color; }
+```
+
+</details>
+
+Change the config and the whole shader — struct fields, binding layout, and generated
+expressions — updates in lockstep. For a fully worked example that adds categorical color tables,
```
froyo-np: just a note on wording - "updates in lockstep" makes it sound like mutation occurs when you make a change - the reality is that you call toSource on a new object and you get a new string back - its a pure function!

site/src/content/docs/webgpu/shader-builder.mdx
```diff
@@ -0,0 +1,365 @@
...
+type PointConfig = {
+    /** Name of the vec2 position column. */
+    position: string;
+    /** Quantitative columns, each gets a vertex attribute and a `<col>_range` uniform. */
+    quantitative: string[];
+};
+
+function buildPointShader(cfg: PointConfig): WgslShader {
+    const Vertex = $d.struct('Vertex', [
+        $d.member('vIndex', $t.u32, [$a.builtin('vertex_index')]),
```
froyo-np: is there any chance I could convince you to drop the cruft on some of these names? 
I think `__brand` could be `brand`, `__gen()` can be `gen()` and `$d` could just be `d` (or something else).

IMO, typing up this stuff is already much less pleasant than just typing WGSL (which is saying something) the symbols add visual clutter, are harder to type - the $ is a particular offender, as its clearly intended to be used very frequently.

froyo-np: `decl`
`attr`
`type`

maybe?

site/src/content/docs/webgpu/shader-builder.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+</details>
+
+Because `Camera` and `VsOut` are ordinary values, you can pass them anywhere a type identifier is
+accepted — as the uniform's type, as a function parameter type, or as an entry point's return
+type — and the generator emits the struct's name.
+
+## Level 3 — Declarative vertex inputs
+
+Buffer-backed vertex attributes have rules: every leaf needs exactly one `@location` or
```
froyo-np, line 197: there are similar rules for storage buffers and uniform buffers (mostly around alignment) - are those rules validated?

site/src/content/docs/webgpu/shader-builder.mdx
```diff
@@ -0,0 +1,365 @@
...
+
+## Level 4 — Textures and samplers
+
+Sampled resources follow the same pattern as uniforms: declare them with a group/binding and use
+them by name in a function body.
+
+```ts
+import { $d, $a, $t, shader, asSource } from '@alleninstitute/vis-core';
+
+const albedo = $d.texture('albedo', $t.texture('2d', 'f32'), 0, 0);
```
froyo-np: this part feels inconsistent -
why does $t.texture use a string for its type instead of $t.f32, and if we are inside the scope of $d.texture, why do I need to re-specify that the type is $t.texture ? what happens if I put a different type in here?
maybe it might make more sense to merge d.sampler and d.texture into something like d.var(name,type) ?

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Checking for alignment in name usage _might_ be possible, but only by actually interpreting the function bodies, sadly.

Which kind of type-correctness are you thinking of? Within-function-body correctness (that would be a no)? Input value correctness (that would depend on other modules in the larger renderer package)?

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Yeah, I was on the fence about even including that chunk of comment. Guess I should've listened to my gut 😛

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Chris commented on the shortened bucket names (`$d`, `$a`, and `$t`) as well 😝 Yeah, I'm leaning towards renaming them, though I can't use `type` since that's a reserved word. How about `decls`, `attrs`, and `types`?

I'd actually much prefer to have the package setup so that devs can import these modules directly and rename them as they choose, but that was going to involve some more complicated bundling setup. Might be worth it, though, for DX.

Jarbuckle: As for the `__x` names -- yeah, I can, I just really hate having "internal use only" fields be named/treated as though they're part of the public "API" of a type. Those are really not meant to be used by developers, though I suppose `gen` wouldn't be too bad to make available.

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: True, though aren't those rules pretty much entirely invisible within the context of the WGSL code? There's code for handling alignment and such in other modules (the ones that actualize the relationship between buffers and uniforms), but it seems sorta out-of-scope here, whereas having multiple `@location`s pointing at the same place or having both `@location` and `@builtin` on a single property are technically incorrect WGSL semantics and would (presumably) cause a compilation error for the shader itself.

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Ahhh, hmm, yes, good catch! Worth a reconsideration, for sure. I noticed the oddity of the `2d` and `f32` parts being strings, compared to the other explicitly-used types, and had it down as a future improvement for making these into proper enums potentially (since these are syntactically specific to the setup of Textures, rather than being "general" types). I shall ponder some revisions here!

site/src/content/docs/webgpu/shader-builder.mdx
froyo-np: hmmm - to me the __brand or __identType is very much a public thing! its kinda the bridge between nominal typing you'd get with classes you can extend and the structural typing we're doing in this library.

I could use __identType to write my own code-generator, or apply validation logic that might be specific to my end-user-application.

site/src/content/docs/webgpu/shader-builder.mdx
froyo-np: I guess what I am getting at is documentation - type-checking was an example.
a section like "Limititations" with a list of common ways someone might generate a shader that will fail to compile / run.

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Okay, that's fair. I've gone ahead and updated the naming conventions for `__gen`, `__brand` and `__identType`.

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Updated! Hopefully it reads more correctly now

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: Hmm, okay; I'll ponder that.

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: I also updated `$a` to `attrs`, `$d` to `decls`, and `$t` to `types`, and added a section in the docs explaining the different ways of importing those constructor functions and the buckets (including aliasing for shorter names, when desired).

site/src/content/docs/webgpu/shader-builder.mdx
Jarbuckle: This has been updated now

PR #285, opened 2026-08-03
webgpu-filtering

froyo-np:
# WebGPU powered filtering

# What

Add a utility to generate semi-typesafe webGPU filtering progams, implemented as compute shaders, with generated, typesafe wrapper functions by which to invoke the filter

```
const filter = given({
        cells: { subclass: 'u32', gene_x: 'f32' },
        edges: { start: 'u32', end: 'u32', str: 'f32' }
    })
      .from('edges')
      .select('cells[start].gene_x') // typesafe - expression must reference a given table.field
      .where('cells[start].subclass == fromClass') // note that the type-safety is fragile - it will break if you omit the whitespace!
      .and('cells[end].subclass == toClass')
      .build()
      .run(
       parameters: {fromClass: 4, toClass: 3},
       /*** more args... ****/
)
```

the above snippet generates a compute shader & pipeline that performs the given predicate on each row, and accumulates the selected data for each row that passes. The invocation of the pipeline is also typesafe*.

Caveats:
* the expressions in where, and and or clauses are typesafe, but weakly so!
  * whitespace must be exact
  * expressions of the form `table[index_expr].field` cannot be recursive (`index_expr` must be a literal field of the from table)
  * the parameter (fromClass and toClass in the example above) must be the righthand side of the predicate
  * the system will not stop you from using an invalid variable name that will fail to compile, eg a predicate like "start == 76^whatever" wont trigger the typecheck failure, despite being invalid WGSL
  * the type of the parameter (on the right) will be inferred as an exact match of the expected type on the left, so if the field is a vector, then so must be the parameter its compared with.
* select can only select a single field of the from table, or the index of the passing row (via '$index')
* you can re-use a parameter, however if you compare it to fields of differing types, its inferred type could become never, or the union of the compared types, neither of which will be helpful to you.
* the results of a run will write to whatever gpuBuffer is bound to the results binding. the values within will be contiguous, that is to say, if 50 randomly distributed rows pass out of 100, all 50 passing selected rows will be at the start of the written buffer.
* a counter buffer is also bound, and will contain the count of passing values after the run.
* the results should not be expected to have any predictable order whatsoever.
* early benchmarking suggests that a wimpy-ish laptop can filter about 1 million rows in about 4 milliseconds.
* `== <= >= > < !=` are supported for scalar types.
* all scalar operators are supported for vector comparisons when prefixed by `any` or `all` eg. `some_vec all(>) param`
* **The most potentially annoying issue**: WGSL does not permit an expression like (A && B || C) - this is because they chose not to implicitly make either operator more "sticky" than the other. the result of this that matters for this tool is that it makes the api kinda gross if you need an expression with both AND and OR - you must parenthesize your predicate expressions manually - `where(A).andOpen(B).or(C).close().and(D)` would result in correct WGSL: `A && (B || C) && D`
* note also - it is quite easy to still generate invalid WGSL - the benefit here is that it is _possible_ (unfortunately not _super easy_) to create a working complex boolean expression
# How

the generated compute shader uses a shared workgroup-local array to collect passing results with minimal contention between threads. then the primary thread (thread 0) of each workgroup adds its count via an atomicAdd, signifying its request for that many values in the globally accessible results array, after which its safe to copy the passing values to that range.

# Screenshots

# PR Checklist


packages/core/src/rendering/webgpu/filter/query.ts
Jarbuckle, line 1: Why is this file called `less`? 🤔

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -0,0 +1,291 @@
...
+    type FT,
+    type ITable,
+    type OP,
+    type PredLst,
+    type Sel,
+    type SimplePredExpr,
+    type Tables,
+    type VOP,
+} from './types';
+import { keys, map, reduce, values } from 'lodash';
```
Jarbuckle: Since these are capabilities built-in to JS/TS arrays, is there a real need to import them from lodash?

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -0,0 +1,291 @@
...
+function fieldType(s: string, table: ITable) {
+    return table[s];
+}
+export function indexExprType(s: string, tables: Tables, from: string) {
+    const ext = looksLikeIndexExpr(s, tables, from);
+    return ext ? ext.type : undefined;
+}
+export function looksLikeIndexExpr(s: string, tables: Tables, from: string) {
+    // if(s.match(/[a].[b]\[.\]\.$0/))
+    // todo: i am on a plane and I dont remember regex and I dont wanna buy wifi
```
Jarbuckle: 🤣 

BEST. TODO. EVER!

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -0,0 +1,291 @@
...
+            }
+        }
+    }
+    return false;
+}
+function genRef(ctx: FilterShaderQueryContext<Tables>, operand: string, indexing: string = 'element') {
+    const indexed = looksLikeIndexExpr(operand, ctx.tables, ctx.from);
+    if (operand === '$index') {
+        return indexing;
+    }
```
Jarbuckle, line 60: Very minor nitpick, but this could be at the start of the function, yes? Might be very-slightly more efficient

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -0,0 +1,291 @@
...
+        case ')':
+            return ')';
+        case 'and':
+            return `&& ${genPred(ctx, e.pred)}`;
+        case 'or':
+            return `|| ${genPred(ctx, e.pred)}`;
+    }
+}
+export function generatePredicateExpr(ctx: FilterShaderQueryContext<Tables>, exprs: [SimplePredExpr, ...PredLst]) {
+    // return exprs.map(e => typeof e === 'string' ? genPred(ctx, e) : `${e.OP === 'and' ? ' && ' : ' || '} ${genPred(ctx, e.pred)}`).join('\n');
```
Jarbuckle: Probably don't need to keep around these commented-out chunks, yeah?

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -0,0 +1,291 @@
...
+    return exprs.map((e) => (typeof e === 'string' ? genPred(ctx, e) : handlePredElem(ctx, e))).join('\n');
+}
+function extractPred(p: SimplePredExpr | Elem<PredLst>): SimplePredExpr | undefined {
+    return typeof p === 'string' ? p : 'pred' in p ? p.pred : undefined;
+}
+export function genUniformParameterStruct(tables: Tables, from: string, exprs: [SimplePredExpr, ...PredLst]) {
```
Jarbuckle, line 100: ...I'm not sure I knew that it was possible to define Array types using rest "parameters" -- cool!

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -0,0 +1,291 @@
...
+}) {
+    const { inputBindings, predicateExpr, uniformStruct, selections, workgroupSize, indexed } = params;
+    // const selectExpr = select.expr === '$index' ? 'tmp - 1' :
+    //   `${select.expr}[tmp - 1]`
+    // const selectExpr = select.expr;
+    const { structName, decl, construct } = generateOutputStructure(selections);
+    const host = /*wgsl*/ `
+
+    ${uniformStruct.decl}
+    ${decl}
```
Jarbuckle: Strictly for readability, this might be nice to have named `outputDecl` or similar. It confused me for a moment because I thought it might be the same `decl` defined in `uniformStruct`.

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -0,0 +1,403 @@
...
+    TsType,
+    FT,
+    Sel,
+} from './types';
+import * as wgh from 'webgpu-utils';
+
+// so, because and and or cannot be mixed without parens in WGSL
+// we need a structure to provide grouping
+// there are a lot of ways we could do this, but I think the easiest
+// that isnt too limiting is CNF, which means the AND of a bunch of clauses,
```
Jarbuckle: I learned something today! Didn't know there was a formal name for that (x | y) & (a | b) & (d | e | f) logical structure -- sweet!

packages/core/src/rendering/webgpu/filter/types.ts
```diff
@@ -0,0 +1,243 @@
+import * as wgh from 'webgpu-utils';
+type LT = 'f32' | 'u32' | 'i32';
+type Abbr = 'f' | 'u' | 'i';
+type Short = 2 | 3 | 4;
+// type VT = `vec${Short}<${LT}>`
+type VT<S extends Short> = `vec${S}${Abbr}`;
```
Jarbuckle: `Short` seems an odd name for this; why not `VLen` or similar?

site/src/examples/filter/filter.ts
```diff
@@ -0,0 +1,152 @@
...
+        .from('edges')
+        .select('$index')
+        .select('cells[start].gene_x')
+        .select('cells[start].subclass')
+        .select('cells[end].subclass')
+        .where('cells[end].subclass == toClass')
+        .andOpen('cells[start].subclass == fromClass')
+        .and('cells[start].position all(>=) minCorner')
+        .and('cells[start].position all(<) maxCorner')
+        .close()
```
Jarbuckle, line 69: Am I right in thinking that `.andOpen() / .close()` effectively surrounds the calls in between with parentheses? In that case, is it meaningful in this case, since it's basically `x & (y & z)`?

Jarbuckle (commented):
Haven't had a chance to run the demo yet due to a module packaging error relating to Lodash, but I've looked over the code and have a few comments. That said, this is a really cool feature and one I'm excited to see us use!

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: haha oh yeah...

my first draft of this whole idea was called "fancy"
this version is the "less" fancy version.

I'll rename it

packages/core/src/rendering/webgpu/filter/gen.ts
froyo-np: ah indeed thanks

packages/core/src/rendering/webgpu/filter/gen.ts
froyo-np: great point, will improve naming

packages/core/src/rendering/webgpu/filter/types.ts
froyo-np: I guess I think of these as "short vectors"... I could be convinced that VLen is good here...

site/src/examples/filter/filter.ts
froyo-np: exactly

packages/core/src/rendering/webgpu/filter/gen.ts
froyo-np: sorry - this is my semi-bad habit from ye olden tymes. I should use the built-in ones...

Jarbuckle (approved):
Example page works!

PR #287, opened 2026-08-13
Noah/refactor filtering

froyo-np:
# What

the initial implementation of filtering did not support some features (like referring to the .x or .y value of a vector column) that were critical for the actual filters we need in our use case. Making this all work required a bit of a rewrite, but its much more capable now, and the "andOpen" style syntax, which was confusing, is now replaced with something that is still a bit confusing, but better and you get more power from it.

# How

use a builder-style expression system, rather than very fragile typescript template-string types.

# Screenshots

# PR Checklist

### DIFF
--- a/packages/core/src/rendering/webgpu/filter/gen.ts
+++ b/packages/core/src/rendering/webgpu/filter/gen.ts
@@ -1,6 +1,6 @@
 import { type WgslType, type Sel, type Tables } from './types';
 
-const entries = (r: object)=> Object.entries(r);
+const entries = (r: object) => Object.entries(r);
 
 function generateOutputStructure(selections: ReadonlyArray<Sel>) {
     // the names of the values in the structure dont matter at all -
--- a/packages/core/src/rendering/webgpu/filter/gen.ts
+++ b/packages/core/src/rendering/webgpu/filter/gen.ts
@@ -6,26 +6,26 @@ function generateOutputStructure(selections: ReadonlyArray<Sel>) {
     // the names of the values in the structure dont matter at all -
     const structName = `OutputStruct`;
     const fields = selections.map((s, i) => `field_${i.toFixed(0)}: ${s.type},`).join('\n');
-    const decl = `struct ${structName} {\n ${fields} \n };`;
+    const structDecl = `struct ${structName} {\n ${fields} \n };`;
     const initializers = selections.map((s) => s.selection).join(', ');
     const construct = `${structName}(${initializers})`;
 
-    return { structName, decl, construct };
+    return { structName,  structDecl, construct };
 }
 export function generateShader(params: {
     workgroupSize: number;
     inputBindings: string;
     predicateExpr: string;
-    uniformStruct: { name: string; typeName: string; decl: string };
+    uniformStruct: { name: string; typeName: string; structDecl: string };
     indexed: boolean;
     selections: ReadonlyArray<Sel>;
 }) {
     const { inputBindings, predicateExpr, uniformStruct, selections, workgroupSize, indexed } = params;
-    const { structName, decl, construct } = generateOutputStructure(selections);
+    const { structName, structDecl, construct } = generateOutputStructure(selections);
     const host = /*wgsl*/ `
 
-    ${uniformStruct.decl}
-    ${decl}
+    ${uniformStruct.structDecl}
+    ${structDecl}
     var<workgroup> results: array<u32,${workgroupSize}>;
     var<workgroup> count: atomic<u32>;
 
@@ -129,7 +129,7 @@ export function assembleQuery<Ts extends Tables>(
             predicateExpr: predExpr,
             indexed,
             selections: ctx.selections.map((s) => ({ selection: s.selection, type: s.type })),
-            uniformStruct: { name: ctx.uniformName, typeName: ctx.uniformTypeName, decl: uniStructDecl },
+            uniformStruct: { name: ctx.uniformName, typeName: ctx.uniformTypeName, structDecl: uniStructDecl },
         }),
         bindingLookups,
     };
--- a/packages/core/src/rendering/webgpu/filter/gen.ts
+++ b/packages/core/src/rendering/webgpu/filter/gen.ts
@@ -1,5 +1,9 @@
 import { type WgslType, type Sel, type Tables } from './types';
+
+// lodash is being really troublesome.. so I wrote my own here
+/* oxlint-disable no-console, typescript/no-explicit-any*/
 const entries = <T extends {}>(r: T): ReadonlyArray<[keyof T, T[keyof T]]> => Object.entries(r) as any;
+
 function generateOutputStructure(selections: ReadonlyArray<Sel>) {
     // the names of the values in the structure dont matter at all -
     const structName = `OutputStruct`;
--- a/packages/core/src/rendering/webgpu/filter/gen.ts
+++ b/packages/core/src/rendering/webgpu/filter/gen.ts
@@ -1,179 +1,34 @@
-// generate the interesting bits of the filter-shader
-
 import {
-    type Elem,
-    type FilterShaderQueryContext,
     type WgslType,
-    type ITable,
-    type OP,
-    type PredLst,
     type Sel,
-    type SimplePredExpr,
     type Tables,
-    type VOP,
 } from './types';
-
-function isVecOp(s: OP | VOP): s is VOP {
-    return s.startsWith('a');
-}
-function parseVecOp(op: VOP) {
-    const aggregation = op.substring(0, 3);
-    const sop = op.substring(4).split(')')[0];
-    return [aggregation, sop] as ['any' | 'all', OP];
-}
-
-// dont export this - its only legit if we know a bunch of stuff about the string
-function fieldType(s: string, table: ITable) {
-    return table[s];
-}
-export function indexExprType(s: string, tables: Tables, from: string) {
-    const ext = looksLikeIndexExpr(s, tables, from);
-    return ext ? ext.type : undefined;
-}
-export function looksLikeIndexExpr(s: string, tables: Tables, from: string) {
-    // the goal here is to parse an expr that looks like:
-    // someTable[someColumn].someOtherColumn
-    // could I have used Regex? yes, and perhaps that would be more elegant!
-    // for now, this works, and its not too long
-
-    const [tbl, rest] = s.split('[');
-    if (tbl && rest) {
-        const [index_field, selection] = rest.split('].');
-        if (index_field && selection) {
-            if (tables[tbl] && tables[from] && tables[from][index_field] && tables[tbl][selection]) {
-                // return the expected type
-                return {
-                    fTable: tbl,
-                    selection,
-                    index_field,
-                    from,
-                    type: tables[tbl][selection],
-                };
-            }
-        }
-    }
-    return false;
-}
-function genRef(ctx: FilterShaderQueryContext<Tables>, operand: string, indexing: string = 'element') {
-    if (operand === '$index') {
-        return indexing;
-    }
-    const indexed = looksLikeIndexExpr(operand, ctx.tables, ctx.from);
-    if (indexed) {
-        return `${indexed.fTable}_${indexed.selection}[${indexed.from}_${indexed.index_field}[${indexing}]]`;
-    } else if (operand in ctx.tables[ctx.from]!) {
-        return `${ctx.from}_${operand}[${indexing}]`;
-    }
-    return operand;
-}
-function genPred(ctx: FilterShaderQueryContext<Tables>, p: SimplePredExpr) {
-    const [lhs, op, rhs] = p.split(' ') as [string, OP | VOP, string];
-    // TODO! handle non-PARAMETER rhs exprs
-    const param = `${ctx.uniformName}.${rhs}`;
-    if (isVecOp(op)) {
-        const [agg, sop] = parseVecOp(op);
-        const str = `${agg}(${genRef(ctx, lhs)} ${sop} ${param})`; // any / all are built-in wgsl fns over vectors of booleans
-        return str;
-    }
-    return `${genRef(ctx, lhs)} ${op} ${param}`;
-}
-function handlePredElem(ctx: FilterShaderQueryContext<Tables>, e: Elem<PredLst>) {
-    switch (e.OP) {
-        case 'and (':
-            return `&& (${genPred(ctx, e.pred)}`;
-        case 'or (':
-            return `|| (${genPred(ctx, e.pred)}`;
-        case ')':
-            return ')';
-        case 'and':
-            return `&& ${genPred(ctx, e.pred)}`;
-        case 'or':
-            return `|| ${genPred(ctx, e.pred)}`;
-    }
-}
-export function generatePredicateExpr(ctx: FilterShaderQueryContext<Tables>, exprs: [SimplePredExpr, ...PredLst]) {
-    return exprs.map((e) => (typeof e === 'string' ? genPred(ctx, e) : handlePredElem(ctx, e))).join('\n');
-}
-function extractPred(p: SimplePredExpr | Elem<PredLst>): SimplePredExpr | undefined {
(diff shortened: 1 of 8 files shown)


packages/core/src/rendering/webgpu/filter/types.ts
```diff
@@ -32,90 +31,129 @@ type alpha =
...
 export const OPS = ['==', '>', '>=', '<', '<=', '!='] as const;
 export type OP = Elem<typeof OPS>;
 // export const VOPS = ['==' , '!=' , 'any <' , 'any >' , 'all <' , 'all <']
 export type VOP = `${'any' | 'all'}(${OP})`;
 
 export type FType<T extends ITable, K extends keyof T> = T[K];
-export type VarName = `${letter}${string}`;
-// type Var<T extends VarName> = `$${T}`
-// I want a type error if the op type is a vOp and either operand is not...
-export type PredicateExpr<T extends ITable, K extends keyof T, Param extends VarName> = K extends string
```
froyo-np, line 56: I'd probably read this code in editor, rather than diff-view - its essentially a re-write of most of the fun parts.

1. gone are the complicated template-string style types.
2. replaced are shorter, templated builder-style functions, which infer their types from the parameters, which are usually very short strings.

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -169,11 +19,11 @@ export function generateShader(params: {
     selections: ReadonlyArray<Sel>;
 }) {
     const { inputBindings, predicateExpr, uniformStruct, selections, workgroupSize, indexed } = params;
-    const { structName, decl: outputStructDecl, construct } = generateOutputStructure(selections);
+    const { structName, decl, construct } = generateOutputStructure(selections);
     const host = /*wgsl*/ `
 
     ${uniformStruct.decl}
-    ${outputStructDecl}
+    ${decl}
```
Jarbuckle: Looks like this got reverted somehow 😅

site/src/content/docs/webgpu/filtering.mdx
```diff
@@ -8,20 +8,21 @@ and a table of connections between those cells (given by indexes start and end),
 the shader is generated using a short, SQL-like builder-style api:
 
 ```ts
-.given({
+const { table, clause, select, all } = given({
```
lanesawyer, line 11: Yeah, I like this approach a bit better. Definitely more verbose, but not detrimentally so.

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
-    > {
-        return new FilterTable<
-            Ts,
-            T,
-            typeof pred extends PredicateExpr<T, Ti, Param>
-                ? Params & { [k in Param]: TsType<T[Ti]> }
-                : Params & { [k in Param]: TsType<Ts[O][F]> }
-        >(this.ctx, [...this.predicates, { OP: 'and', pred }]);
+/* oxlint-disable no-console, typescript/no-explicit-any*/
+const entries = <T extends {}>(r: T): ReadonlyArray<[keyof T, T[keyof T]]> => Object.entries(r) as any;
```
lanesawyer: Hm, I don't love the `as any` force casting here and in `gen.ts` but I'm not seeing a really straightforward other option that I've liked as I've played aroudn with it.

site/src/examples/filter/filter.ts
```diff
@@ -52,22 +52,65 @@ function generateFakeDataset(device: GPUDevice, edges: number, cells: number): g
...
-        .select('cells[end].subclass')
-        .where('cells[end].subclass == toClass')
-        .andOpen('cells[start].subclass == fromClass')
-        .and('cells[start].position all(>=) minCorner')
-        .and('cells[start].position all(<) maxCorner')
-        .close()
-        .build(device, 'testing');
+    const { all, any, column, table, select, clause } = given(tableLayout).from('edges');
+
+    const filter = select('$index')
```
lanesawyer, line 61: Demo still works and shows a table of read data!

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+                        uniformTypeName: 'Params',
+                    },
+                    predicateExpr,
+                    paramDecls,
+                    64,
+                    false
+                );
+                return Q.shader;
+            },
+            build: (device: GPUDevice, label: string) => {
```
lanesawyer, line 451: `build` and `buildIndex` are _very_ similar, maybe extract a helper function? Or do we expect them to diverge more substantially over time?

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+                    };
+                    (runner as (args: RunFilterArgs<Ts>) => void)(args);
+                }
+                enc.copyBufferToBuffer(results, resolve);
+                enc.copyBufferToBuffer(resultsCounter, resolveCount);
+                dev.queue.submit([enc.finish()]);
+                // ok now get the results and compare them!
+                Promise.all([resolve.mapAsync(GPUMapMode.READ), resolveCount.mapAsync(GPUMapMode.READ)])
+                    .then(() => {
+                        const count = new Uint32Array(resolveCount.getMappedRange());
```
lanesawyer: This variable is unused. Should it be used? Otherwise delete!

packages/core/src/rendering/webgpu/filter/query.ts
lanesawyer, line 1: If we could get some tests for this file in particular, it would give me more confidence in the changes

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
     return entries(tables)
         .flatMap(([name, table]) => {
             return entries(table).map(([field, buffer]) => {
-                const b = lookups[name]?.[field];
+                // I promise, these are all strings, its ok
+                const b = lookups[name as string]?.[field as string];
                 if (b) {
-                    return { resource: buffer, binding: b };
+                    return { resource: buffer as GPUBuffer, binding: b };
                 }
```
lanesawyer: Would be nice to get rid of the `as` and prove it, not too much extra code and doesn't change the logic if it does fail:

```suggestion
                if (typeof name === "string" && typeof field === "string") {
                    const b = lookups[name]?.[field];
                    if (b) {
                        return { resource: buffer as GPUBuffer, binding: b };
                    }
                }
```

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
-        });
+        const START_BINDING = 3;
+        const UNI_INSTANCE_NAME = 'params';
+        const UNI_STRUCT_TYPE_NAME = 'Params';
+        const WG_SIZE = 64;
+        const paramTypes = predicates.paramDeclarations();
+        const paramDecls = entries(paramTypes)
+            .map(([name, type]) => `${name}:${type}`)
+            .join(',\n');
+        // const uniDecl = `struct ${UNI_STRUCT_NAME} {
```
lanesawyer: Can remove commented out code

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+            indexed: Indexed;
+            label: string;
+            wgSize: number;
+            serializeParameters: (p: Params, buffer?: ArrayBuffer) => ArrayBuffer;
+        }) => {
+            const { device, pipe, safeLookups, indexed, label, wgSize } = options;
+            const runner = (args: Indexed extends true ? RunIndexedFilterArgs<Ts> : RunFilterArgs<Ts>) => {
+                const { enc, parameters, sets, timestampWrites } = args;
+                // here, we zero out the result counters
+                // TODO - consider not doing this - if we didnt do that:
```
lanesawyer: A TODO worth solving now?

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+                    : args.sets.map((s, i) => {
+                          return device.createBindGroup({
+                              layout: pipe.pipeline.getBindGroupLayout(1),
+                              entries: [
+                                  { binding: 0, resource: s.resultCounter },
+                                  { binding: 1, resource: s.results },
+                                  ...mapTablesToBindings(s.tables, safeLookups),
+                              ],
+                          });
+                      });
```
lanesawyer, line 265: Could simplify this a bit, the only difference is the `{ binding: 2, resource: s.elements },` bit. Don't know that I love having the `indexed && "elements" in s` check though, could obfuscate a little bit what's happening, since folks might not look deeply into `entries` and not understand the `indexed` versus not `indexed` take.

Up to you, just saw a little bit of code duplication that could be avoided in a few different ways!

```typescript
                const bindings = args.sets.map((s) =>
                    device.createBindGroup({
                        layout: pipe.pipeline.getBindGroupLayout(1),
                        entries: [
                            { binding: 0, resource: s.resultCounter },
                            { binding: 1, resource: s.results },
                            ...(indexed && "elements" in s ? [{ binding: 2, resource: s.elements }] : []),
                            ...mapTablesToBindings(s.tables, safeLookups),
                        ],
                    }),
                );
```

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+                              timestampWrites,
+                          }
+                        : { label }
+                );
+                pass.setPipeline(pipe.pipeline);
+                pass.setBindGroup(0, bg0);
+                for (let i = 0; i < sets.length; i++) {
+                    const s = sets[i]!;
+                    const bg1 = bindings[i]!;
+                    // console.log('running filter-->',s.results.label)
```
lanesawyer, line 285: Can remove, or potentially do a `logger.debug` here so we could turn these on later if we choose.

I don't think our logger gets stripped from the build though so no-ops would be annoying for performance...  Maybe something for us to look into later so we can feel more free to sprinkle debug logs around.

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+                const tableBuffers: BufferTables<Ts> = mapValues(
+                    tables,
+                    (table: Record<string, { buffer: ArrayBuffer }>) =>
+                        mapValues(table, (column: { buffer: ArrayBuffer }) => {
+                            const buf = dev.createBuffer({ size: column.buffer.byteLength, usage: R });
+                            dev.queue.writeBuffer(buf, 0, column.buffer);
+                            inputs.push(buf);
+                            return buf;
+                        })
+                ) as BufferTables<Ts>; // TS cant tell, mapValues I think does erase the info...
```
lanesawyer, line 321: `mapValues` does have a generic parameter but I'm having trouble making it happy...

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,566 @@
...
+    ScalarType,
+    VLen,
+    VectorType,
+    onlyLetters,
+    FType,
+    Cmp,
+    ArrayBufferTables,
 } from './types';
-import * as wgh from 'webgpu-utils';
+import * as lo from 'lodash';
```
lanesawyer, line 30: Oooh maybe we move to `lodash-es` here too soon, didn't realize we had only done that in one repo so far.

lanesawyer (commented):
Handful of comments, looks to be working well, the demo is still functional!

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: yeah I'd love to - ye olde lodash imports break the build regularly unless I do absolutely wacky stuff.

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: I like it!

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: they differ only in that one returns a function which requires an `elements` buffer to be passed, full of indexes. I think I can wrestle the typechecker... these are only separate so that the inferrence of that requirement works out - I'll give it a shot.

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: yeah - the way it works out is essentially `Object.entries(r) as any as ReadonlyArray<[keyof T, T[keyof T]]>`
mostly this was me struggling with both lodash entries and Object.entries types given my very fragile type stuff going on here - I'll try and get rid of the cast, but to be fair, its pretty contained - the any cant escape!

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: ok - so the entries types and the keyof typeof T[K]TS[K]T wacky fun that I was doing made me go bananas - this whole situation is solved by not messing with object.entries - all the places at runtime (like here) dont care about the types, just the strings.

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: ok - this was noah having Typescript fever - I've removed the nonsense here and now lots of stuff lower down no longer needs silly casting.

packages/core/src/rendering/webgpu/filter/query.ts
Jarbuckle: Yeah, also dug into this one, not finding a working variant of `mapValues` that solves this problem -- all of them erase key-typing in one way or another. 😞

packages/core/src/rendering/webgpu/filter/query.ts
```diff
@@ -1,387 +1,523 @@
...
+                        for (let i = 0; i < expected.buffer.byteLength; i++) {
+                            if (dv.getUint8(i) !== ex.getUint8(i)) {
+                                logger.error('filter validation failed at byte: ', i);
+                                failBytes += 1;
+                            }
+                        }
+                        if (failBytes === 0) {
+                            logger.info('validation success');
+                        }
+                    })
```
Jarbuckle, line 380: How many comparisons is this doing, on average? 🤔 And do we need to know the number of bytes that failed? Also, genuinely curious if we want to keep going if we get failures? Would that be a filter algorithm issue, a hardware issue, both, neither?

Jarbuckle (commented):
Just a couple things, looking good overall!

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: the idea of this is a smoke-test. it will certainly fail if you ask it to filter more than 64 rows of data, so at worst, this will fail after 512 checks. I've re-written this to return a copy of the result buffer in the failing case, and it now stops at the first failed byte.

packages/core/src/rendering/webgpu/filter/gen.ts
froyo-np: fixed, sorry about that...

packages/core/src/rendering/webgpu/filter/query.ts
froyo-np: tests written!

packages/core/src/rendering/webgpu/filter/query.test.ts
lanesawyer, line 1: Simple but effective at proving the output looks as expected! Thanks!

packages/core/src/rendering/webgpu/filter/query.ts
lanesawyer: Replacement PR here! https://github.com/AllenInstitute/vis/pull/288

We can merge yours first and I can come clean up the remainder later, or you can review and merge that and start using in this PR. Either way!

lanesawyer (approved):
Changes make sesne, thanks for the test, and the demo still runs. I think any other pain points can be ironed out again as we hit them while making further progress on Connectomics!

PR #293, opened 2026-08-25
feat: share 2D camera math and OME-Zarr render channels [DT-9938]

lanesawyer:
# What

Moves 2D camera math from other apps into `vis-geometry` and OME-Zarr render channel construction into `vis-omezarr`, and uses those new helper functions in some of the examples, which actually fixes a couple bugs in our example site!

Both migrations came out of looking into [DT-9938](https://alleninstitute.atlassian.net/browse/DT-9938), where a Neuroglancer preview in `bkp-client` rendered noticeably dimmer than Neuroglancer itself. There are a handful of the same or similar logic scattered across `bkp-client`, and centralizing them all is a helpful step.

# How

**`vis-geometry/camera.ts`** — `screenToData`, `zoomAround`, `zoom`, `pan`, `fitToScreen`, exported from the barrel as a `Camera2D` namespace so these generic names stay unambiguous at import sites.

`bkp-client` had three copies of this math (its PPDC web component, its React OME-Zarr preview, and its Connectomics slice), and they had drifted apart on which way the vertical axis runs. One viewer zoomed toward a vertically mirrored point, disagreeing with its own pan code.

This started in core, on the reasoning that it encodes a rendering convention rather than pure math. Noah asked for it in geometry, so that is where it lives now. It only ever needed `Box2D` and `Vec2`, so geometry takes on no new dependencies, and since geometry depends on nothing there is no cycle.

**`vis-omezarr/render-channels.ts`** — `renderChannelsFromMetadata`, `fallbackRGBChannels`, `clampGamutToDataRange`.

`colorChannels` maps omero's `window.start/end` to `range` and `window.min/max` to `window`, then hands out an unclamped `range` that consumers naturally use as a gamut. Files in the wild declare a display range past what the data can hold (e.g. the file behind DT-9938 describes uint8 data with a blue channel ending at 473) so that gamut spreads across values the data never reaches and the channel renders dim.

`renderChannelsFromMetadata` clamps to the data range, which is what the tools writing these files already do when they configure Neuroglancer.

Also exports `OmeZarrColorChannel`. It was exported from `zarr/types.ts` but never added to the package barrel, so consumers could not name the type.

# PR Checklist

- [x] Have you smoke tested the example applications?
- [x] Did you check that the changes meet accessibility standards? (no UI)
- [x] Have you tested the application on these browsers?

🤖 Generated with [Claude Code](https://claude.com/claude-code), cleaned up by a human

[DT-9938]: https://alleninstitute.atlassian.net/browse/DT-9938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


site/src/examples/omezarr/minimal-example/sliceview.tsx
lanesawyer, line 1: This example isn't rendered in the example site, but I replaced the functions here anyway. Could delete this or expose it on the Examples page, either as part of this PR or another time.

packages/omezarr/src/index.ts
```diff
@@ -21,6 +22,7 @@ export {
     type OmeZarrOmeroChannelWindow,
     type OmeZarrOmeroChannel,
     type OmeZarrOmero,
+    type OmeZarrColorChannel,
```
lanesawyer, line 25: Should've been exporting this originally!

site/src/examples/common/camera.ts
```diff
@@ -1,43 +1,8 @@
-import { Box2D, Vec2, type box2D, type vec2 } from '@alleninstitute/vis-geometry';
+import type { box2D, vec2 } from '@alleninstitute/vis-geometry';
 
 // a basic camera, for viewing slices
 export type Camera = {
```
lanesawyer, line 4: Could move this type elsewhere and delete this file entirely... 🤷🏻

packages/omezarr/src/sliceview/render-channels.ts
```diff
@@ -0,0 +1,57 @@
+import type { Interval } from '@alleninstitute/vis-geometry';
+import type { OmeZarrColorChannel, OmeZarrMetadata } from '../zarr/types';
+import type { RenderSettingsChannels } from './slice-renderer';
+
+const DEFAULT_GAMUT: Interval = { min: 0, max: 80 };
```
lanesawyer, line 5: This comes from PPDC, not fully sure 80 makes sense to ship here? This _shouldn't_ really be hit all that often though anyway. Open to suggestions!

packages/geometry/src/camera.ts
```diff
@@ -0,0 +1,76 @@
...
+    const minCorner = Vec2.scale(Vec2.sub(imageSize, viewSize), 0.5);
+    return Box2D.create(minCorner, Vec2.add(minCorner, viewSize));
+};
+
+/**
+ * The 2D camera helpers, grouped so that names as generic as `zoom` and `pan`
+ * do not land in this package's flat barrel.
+ *
+ * Exported as an object rather than `export * as Camera2D` because the latter
+ * compiles to a TypeScript `NamespaceExport`, which parcel's .d.ts transformer
```
lanesawyer, line 68: This makes me want to move to Vite+ in `vis` soon and ditch `parcel` (which seems dead, haven't seen anything ship in 7 months).

froyo-np: I'd like to ditch parcel - it makes some very strange decisions sometimes when generating types.d.ts files, particularly with the complicated kind in the webGPU filtering stuff I'm working on

site/src/examples/omezarr/minimal-example/sliceview.tsx
froyo-np: yeah this is very old - we could delete it, fine by me

packages/geometry/src/camera.ts
```diff
@@ -0,0 +1,76 @@
+import { Box2D, Vec2, type box2D, type vec2 } from '@alleninstitute/vis-geometry';
+
+/**
+ * Converts a canvas pixel position to data space.
```
froyo-np, line 5: this file is great overall, but I do feel like its clearly geometry, more than anything else

froyo-np (approved):
I do think this belongs in vis-geometry, rather than vis-core, but either way, lgtm

packages/geometry/src/camera.ts
lanesawyer: Sounds good, I'll swap it over!

PR #301, opened 2026-09-04
fix(vis-core): filter compute shaders incorrectly handled alignment of vec3 columns

froyo-np:
# Fix compute shader handling of vec3 columns

# What
- [x] fixes a potential bug in how vec3 data is handled in webGPU compute shaders - see https://toji.dev/webgpu-best-practices/compute-vertex-data for good background info
- [x] unit tests
# How
- [x] when generating a shader that access vec3 "column" data, instead re-write the shader to access that data as an array of scalars, rather than vectors - such an array magically avoids any alignment problems... accesses of the given column are then re-written to compose vector values manually, using manual indexing calculations.
# Screenshots

# PR Checklist

### DIFF
--- a/packages/core/src/rendering/webgpu/filter/gen.ts
+++ b/packages/core/src/rendering/webgpu/filter/gen.ts
@@ -38,34 +38,37 @@ function parseVecOp(op: VOP) {
     return [aggregation, sop] as ['any' | 'all', OP];
 }
 export function setupExprBuilder(from: string, tables: Tables) {
-    function columnAt(table: string, column: string, indexing: string) {
+    function columnAt(table: string, column: string, indexing: string, swizzle: string | undefined) {
         const type = tables[table][column];
         if (type.startsWith('vec3')) {
             // webGPU can handle vec3 as vertex data (tight 12byte pack) but not in a storage buffer, because why be nice to developers?
             // so we do this silly hack:
+            const offsets: Record<string, string> = { x: '0', y: '1', z: '2' };
+            if (swizzle) {
+                if (swizzle in offsets) {
+                    const offset = offsets[swizzle];
+                    return `${table}_${column}[(${indexing})*3+${offset}]`;
+                }
+            }
             return `${type}(${table}_${column}[(${indexing})*3],${table}_${column}[(${indexing})*3+1],${table}_${column}[(${indexing})*3+2])`;
         }
         // otherwise, be normal about stuff
-        return `${table}_${column}[${indexing}]`;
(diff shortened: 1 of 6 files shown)


packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -15,26 +24,51 @@ function generateOutputStructure(selections: ReadonlyArray<Sel>) {
...
+    return 'i32';
+}
 function parseVecOp(op: VOP) {
     const aggregation = op.substring(0, 3);
     const sop = op.substring(4).split(')')[0];
     return [aggregation, sop] as ['any' | 'all', OP];
 }
-export function setupExprBuilder(from: string) {
+export function setupExprBuilder(from: string, tables: Tables) {
+    function columnAt(table: string, column: string, indexing: string, swizzle: string | undefined) {
```
froyo-np, line 41: here is the important part - 12 byte alignment of data is fine in arrays for vertex buffers, but in storage buffers, the alignment is assumed to be 16 bytes. because of this, we have to interpret arrays of vec3 data as scalars, and do the vector math stuff ourselves. its unclear why WebGPU chooses not to do this for us...

packages/core/src/rendering/webgpu/filter/gen.ts
```diff
@@ -15,26 +24,53 @@ function generateOutputStructure(selections: ReadonlyArray<Sel>) {
...
+        const type = tables[table][column];
+        if (type.startsWith('vec3')) {
+            // webGPU can handle vec3 as vertex data (tight 12byte pack) but not in a storage buffer, because why be nice to developers?
+            // so we do this silly hack:
+            const offsets: Record<string, string> = { x: '0', y: '1', z: '2' };
+            if (swizzle) {
```
Jarbuckle, line 47: Hmm, does swizzling here only support a single accessor, like `A.x` or `A.y`? My understanding of swizzling is that while it technically could include basic indexing, it's really more intended to be used for projections of various kinds, like `A.xy` or `A.zx`. Are those supported?

Jarbuckle (approved):
Just one question in there that doesn't necessarily block anything; otherwise, looking good! Fascinating issue... wonder what the original reason was for such an odd restriction...

packages/core/src/rendering/webgpu/filter/gen.ts
froyo-np: you are completely correct  - most people say 'swizzle' to mean an exotic vector selection like `A.zyx` or what have you. I did initially intend to support such things, but for now, we can only handle simple, single vector component swizzles. This particular case is already limited to vec3's specifically, which is why `w` isn't in the offsets map either.

PR #302, opened 2026-09-09
chore(vis-core): Adding WebGPU Vitest configuration

Jarbuckle:
# chore(vis-core): Adding WebGPU Vitest configuration

# What

Adds a Vitest setup file defining the WebGPU runtime globals (GPUBufferUsage, GPUShaderStage, GPUTextureUsage, GPUColorWrite) and Float16Array, none of which Node provides. The existing filter query path already dereferences GPUBufferUsage, and the WebGPU rendering modules landing in follow-up PRs depend on all of them.

Base of the new WebGPU Rendering System stack; memory, resources, renderer, and the demo page follow.

# PR Checklist


vitest.config.ts
```diff
@@ -1,7 +1,9 @@
+import { fileURLToPath } from 'node:url';
 import { defineConfig } from 'vitest/config';
 
 export default defineConfig({
     test: {
+        setupFiles: [fileURLToPath(new URL('./vitest.setup.ts', import.meta.url))],
```
lanesawyer, line 6: Can this not be just the string?

Docs suggest it'll resolve just fine: https://vitest.dev/guide/learn/setup-teardown.html#setup-files

```suggestion
        setupFiles: ['./vitest.setup.ts']
```

vitest.setup.ts
```diff
@@ -0,0 +1,66 @@
+/**
+ * Vitest global setup. Polyfills typed-array constructors and WebGPU globals that webgpu-utils
+ * references at module-init time but Node hasn't exposed by default yet.
+ *
+ * `Float16Array` is stage-4 ES but in Node 22.x requires the `--js-float16array` flag. We map it
```
lanesawyer, line 5: What about Node 24? Or 26 (which becomes LTS in a few weeks)? Not sure if we can simplify or remove pieces if the newer versions have better support?

lanesawyer (commented):
One simplification and a question

vitest.setup.ts
Jarbuckle: Ahhh, very good callout! It does look like it was released as stable in Node 24, so parts of this may be unnecessary -- sweet!

